PR #12, opened 2020-09-22
CFIN-274 Pattern Library styling

ChristianBeddows:
- Used the React pattern library components lib to add the header, title bar, menu and footer to the results page.
- Created a 'Course' component for the results that we'll be rendering
- Added some styling the 'Course' component to make it more visually similar to the programme catalogue and proposals app
- Clicking on a result will direct the user to the full course page (I don't think this is really in the scope of this task, so if you have an issue with it and would like me to remove this I won't be offended)

Note: from this point onwards, you'll need to setup a .npmrc file in your local copy (see the readme [here](https://github.com/university-of-york/esg-lib-pattern-library-react-components) for some helpful instructions)


src/pages/index.js
```diff
@@ -1,12 +1,25 @@
+import React from "react";
 import PropTypes from "prop-types";
+import {MenuNavigation, UniversityFooter, UniversityHeaderWithSearch, UniversityTitleBar} from "@university-of-york/esg-lib-pattern-library-react-components";
+import {Course} from "../components/Course";
+require("regenerator-runtime/runtime");
```
FergusMcGlynn, line 5: Why is `regenerator-runtime` required?

src/components/Course.module.css
```diff
@@ -0,0 +1,11 @@
+.course {
+    list-style: none;
+    padding: 10px;
+    margin: 0;
+    color: #00b3eb;
```
FergusMcGlynn, line 5: Should we be hard-coding colours like this? I thought we'd be aiming to let the pattern library make those decisions for us?

src/components/Course.module.css
```diff
@@ -0,0 +1,11 @@
...
+}
+
+.course:hover {
+    cursor: pointer;
+    background-color: rgba(0,203,255,.13);
```
FergusMcGlynn, line 10: Same again regarding hard-coded colours

src/pages/index.js
```diff
@@ -1,12 +1,25 @@
...
 
 const App = (props) => {
     return (
         <>
-            <h1>Course search results</h1>
-            {props.searchResults?.map((course) => (
-                <h2 key={course.liveUrl}>{course.title}</h2>
-            ))}
+            <UniversityHeaderWithSearch/>
+            <UniversityTitleBar title="Course search results"/>
```
FergusMcGlynn, line 11: I spun this up locally and noticed that it displays two horizontal black title bars, one with this text in it, and another one that's just blank. On the university homepage, that blank bar contains links to sub-pages, which may or may not be relevant to us. Is there any way of removing it? At least until we have some content to put there?

src/components/Course.js
```diff
@@ -0,0 +1,27 @@
...
+const Course = ({course, router = useRouter()}) => {
+
+    const goToCoursePage = event => {
+        event.preventDefault;
+        router.push(course.liveUrl);
+    };
+
+    return (
+        <li className={styles.course} onClick={goToCoursePage}>
+            <h3>{course.title}</h3>
```
FergusMcGlynn, line 15: Any particular reason for `<h3>`? Just curious about whether `<h1>` and `<h2>` are already defining a structure into which this `<h3>` fits.

src/components/Course.js
```diff
@@ -0,0 +1,27 @@
...
+        </li>
+    );
+};
+
+Course.propTypes = {
+    course: PropTypes.shape({
+        title: PropTypes.string,
+        liveUrl: PropTypes.string,
+    })
+};
```
FergusMcGlynn, line 25: I suspect we are quite quickly going to have this block duplicated between this file and `index.js`. It's not a huge deal - but I wonder if there's a nice way of being able to define this structure once and reuse it in multiple places?

src/components/Course.js
```diff
@@ -0,0 +1,27 @@
+import React from "react";
+import PropTypes from 'prop-types';
+import styles from './Course.module.css';
+import {useRouter} from "next/router";
+
+const Course = ({course, router = useRouter()}) => {
+
+    const goToCoursePage = event => {
+        event.preventDefault;
+        router.push(course.liveUrl);
```
FergusMcGlynn, line 10: Just out of curiosity - what are the benefits of using `router` here rather than just a plain old `<a href=...>`? I'd understand if the user was navigating to a page within the application - but to an external page there doesn't seem to be any particular difference.

src/pages/index.js
```diff
@@ -1,12 +1,25 @@
...
+            <MenuNavigation/>
+
+            <div className="o-wrapper o-wrapper--main o-grid js-wrapper--main">
+                <ul className="course-results">
```
FergusMcGlynn, line 15: Where is the class `course-results` defined?

src/components/Course.js
FergusMcGlynn: Also a `<h3>` within a `<li>` seems a bit odd to me, structurally. Maybe it's fine. I just don't really think of headings existing within a list item.

FergusMcGlynn (commented):
There are a bunch of warnings that get returned when I run the tests - perhaps these could be looked at?

Also XO returns a few errors which ought to be looked at

It's looking good though! :)

FergusMcGlynn (changes requested):
I meant to hit `Request changes` rather than `Comment`... At least for the tests and the XO errors - and it'd be good to resolve some of the other points too if possible.

src/tests/components/Course.test.js
```diff
@@ -0,0 +1,32 @@
...
+        const course = {
+            title: 'Mathematics',
+            liveUrl: courseUrl
+        };
+
+        let router = [];
+
+        render(<Course course={course} router={router}/>);
+
+        fireEvent.click(screen.getByRole('listitem'));
```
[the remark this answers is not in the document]
FergusMcGlynn: Also that article that I linked to is where I read the React Testing Library author asserting that [we should try to use `getByRole` as much as possible](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-byrole-most-of-the-time) when testing.

src/pages/index.js
ChristianBeddows: I left a comment about this as I thought it might lead to some confusion - perhaps I'm using Github wrong, if you didn't see it.

`regenerator-runtime` is required by the pattern library components - I'm not sure why this isn't resolved by the library's own dependencies

src/pages/index.js
ChristianBeddows: Yes, the blank bar is the `MenuNavigation` component. I added it because we tend to use it in most of our apps but perhaps we don't need it. I agree that its worth removing until we actually have options to put in it.

src/pages/index.js
ChristianBeddows: I like to use semantic classnames for certain elements, especially lists and items, but this class isn't used to apply any styling at the moment. Would it be better to remove it?

src/tests/components/Course.test.js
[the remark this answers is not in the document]
ChristianBeddows: `userEvent` sounds pretty good, just from reading the docs, particularly for simulating change events. I don't know if it makes much difference for click events, but I've switched it over anyway

src/components/Course.js
ChristianBeddows: Have done this, will add to the review shortly.

ChristianBeddows (commented):
> There are a bunch of warnings that get returned when I run the tests - perhaps these could be looked at?
> 
> Also XO returns a few errors which ought to be looked at
> 
> It's looking good though! :)

1. Why can't I just reply to this comment as a thread?
2. I've narrowed it down and those warnings are coming from the `UniversityFooter` component from the pattern library, so I don't think we're able to resolve this (short of going and PR'ing a fix into the lib itself).

Will look at the XO errors

src/pages/index.js
FergusMcGlynn: Fair enough - though `course-results` seems more like an `id` than a `class` in DOM terms. Unless you were imagining that elements other than this `<ul>` would also have the class `course-results`?

FergusMcGlynn: Personally I would remove it if it's not being used. I don't have a problem with applying classes (or ids) to elements, but if some elements get class names or ids and others don't then it feels a bit inconsistent, hence why I'd probably opt for omitting them.

src/components/Course.js
ChristianBeddows: Probably none - this function is a result of me wanting to be semantic with my html - I feel that the results should be a list and thus each item in the list should be a `li` rather than an `a`, but perhaps this is adding complexity here

src/components/Course.js
ChristianBeddows: No real reason, I suspect we don't want to use headings at all for the list of results? I feel like this will be updated in the future anyway, so I didn't put too much thought into it. It could just as well be a `p`, but without extra styling that would be much smaller on-screen, so I didn't use that.

FergusMcGlynn: I think headings would be fine for the course titles - but I'm happy to be persuaded otherwise. I think it should probably be `<h2>` rather than `<h3>` though, because the page has a `<h1>` for the page title, but no `<h2>`s beneath that, so it's odd to have `<h3>`s there. Also I've googled whether it's OK to have `<h2>`s inside `<li>`s and it seems to be considered fine by the handful of posts that I found.

src/constants/CourseModel.js
```diff
@@ -0,0 +1,18 @@
...
+    department: PropTypes.string,
+    level: PropTypes.string,
+    length: PropTypes.string,
+    typicalOffer: PropTypes.string,
+    yearOfEntry: PropTypes.string,
+    distanceLearning: PropTypes.string | PropTypes.bool, //This will be a boolean in the future
+    summary: PropTypes.string,
+    imageUrl: PropTypes.string,
+    ucasCode: PropTypes.string,
+});
```
FergusMcGlynn, line 16: This is definitely an improvement on duplicating the course model in `index.js` and `course.js`. I think for this ticket the model only needed `title` and `liveUrl`, but since you've done the rest there's no point removing them. I'm a little uncomfortable with `distanceLearning: PropTypes.string | PropTypes.bool` - I think it should just be one or the other. Even though it'll be changing at some point.

FergusMcGlynn: Actually on second thoughts, I think I might prefer if we remove the fields that we're not currently using. The reason being that we don't yet know what features we're going to implement, or when, and we should probably only be referencing the fields that we are actually using in this model.

src/components/Course.js
ChristianBeddows: `<a>` is indeed much simpler - no need to introduce any added complexity which needs its own tests

src/components/Course.js
ChristianBeddows: I'm just using `<a>` instead of a heading - makes more sense to me! From updating the tests, it feels more semantically correct that we're doing `getByRole('link', {name:"Maths"})` from a user perspective. We can always change this again in the future

src/components/Course.module.css
ChristianBeddows: Following a conversation with Nick, there is no easy way to pull the colour values out of the pattern library. However, we can use the colours in the same way the pattern library does! By having a single constant map defined for all the values and retrieving them in other style declarations using the `colour("blue", 700)` function, we can ensure that we're always using palette colours, and greatly reduce the amount of work needed to update the palette in the future.

.eslintignore
```diff
@@ -0,0 +1 @@
+setupTests.js
```
ChristianBeddows, line 1: XO is throwing an error in this file about jest not being defined. This is false, and any attempt I've made to define/import it causes a different error.

.xo-config.json
```diff
@@ -11,6 +11,7 @@
     "capitalized-comments":"off",
     "indent":"off",
     "import/no-unassigned-import":"off",
+    "import/extensions": "off",
```
ChristianBeddows, line 14: XO was throwing an error because we're now importing a `.scss` module, which is an unrecognised format. I don't think that's a useful error to have (we'll notice whilst developing if there's an issue processing an import) so I've disabled the rule here.

src/constants/styles/colours.scss
```diff
@@ -0,0 +1,29 @@
+$colours:
```
ChristianBeddows, line 1: This is the colour palette as defined by the pattern library (all code in this file copy/pasted from the source code). This allows us to use the `colour` function in our styling to ensure we're always using palette colours, and never hardcoding. We have all of the values in one place, so we can easily update the palette if the library changes in the future.

FergusMcGlynn: This is definitely an improvement over ad-hoc colour definitions in CSS files. I was wondering if there was any way of not having to define colours, or indeed much CSS, at all - and just use pattern library defined classes when trying to style elements? When I look at [the existing Course Finder tool](https://www.york.ac.uk/study/undergraduate/courses/search/?mode=&q=history&level=undergraduate), and look at the CSS for the links and headings, they all come from `https://www.york.ac.uk/static/stable/css/styles.min.css`. I was thinking we should be following that kind of approach, and making use of the University styles so that we don't need to do anything if they ever change. I was thinking we ought to avoid defining our own CSS unless we absolutely need to. What do others think?

ce666-york: I agree with @FergusMcGlynn . This should be a fairly straightforward page so I'd hope it can be implemented without custom components or styling

setupTests.js
```diff
@@ -11,3 +11,7 @@ jestFetchMock.enableMocks();
 require("dotenv").config({
     path: ".env.test",
 });
+
+jest.mock("./src/components/Course.module.scss", () => {
```
ce666-york, line 15: Is this the best place for this? For me ideally it would be closer to the tests being run (even if that means duplication)

src/constants/model/CourseModel.js
```diff
@@ -0,0 +1,18 @@
+import PropTypes from "prop-types";
+
+const COURSE_MODEL = PropTypes.shape({
```
ce666-york, line 3: I'd be tempted to develop incrementally and only include the fields that we need

src/pages/index.js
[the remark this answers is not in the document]
ce666-york: I would love to get this resolved but it sounds quite complicated. I'll catch up on the Slack thread and then maybe we can discuss in more detail?

ce666-york (changes requested):
I'd like us to spend a bit more time thinking about the Course component and seeing if we can avoid custom styling. I'd be happy to pair with someone on it!

src/pages/index.js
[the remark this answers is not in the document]
FergusMcGlynn: Matt Bodey stated ```It's definitely a Babel issue, you absolutely should not have to include regenerator runtime yourself``` which makes me hesitant to persist with `regenerator-runtime` as a project dependency. I guess we'll need to make a judgement about how much time/effort it will be to fix the problem with babel.

PR #14, opened 2020-09-24
Simple Course Search results page using Pattern Library styling

ce666-york:
Following a conversation with @FergusMcGlynn and @RichardAllum we agreed to take a step back and split the work out into designing the page, removing the regenerator-runtime dependency, and resolving the linting errors. 

This branch should provide a baseline to do all of these.


ce666-york:
Running `npm run lint` doesn't report any errors for me so I'm not sure whether there's something I'm missing that needs to be done.

FergusMcGlynn (approved):

PR #15, opened 2020-09-24
CFIN-289: Handle errors when calling Courses API

FergusMcGlynn:


ncn1 (changes requested):
I've checked this out and there is a failing test. Also, I altered the call to the AP I and the error was not handled - though latter might be that the API iitself didn't error, just this app couldn't handle the results

FergusMcGlynn:
> I've checked this out and there is a failing test. Also, I altered the call to the API and the error was not handled - though latter might be that the API itself didn't error, just this app couldn't handle the results

Oh really? Which test is failing? They all pass on my machine...

Good point about the app not handling the error properly - I will look into this

ncn1:
> Oh really? Which test is failing? They all pass on my machine...

Oops - only failing because I forgot to run npm install! Trying to do in a hurry before the refinement session.

FergusMcGlynn:
> Oops - only failing because I forgot to run npm install! Trying to do in a hurry before the refinement session.

No problem - I have updated the code so the app should behave as expected now

src/pages/index.js
```diff
@@ -18,14 +19,32 @@ App.propTypes = {
...
 
-    return { props: { searchResults: data.results } };
+    return { props: { isSuccessfulSearch, searchResults: searchResponseData.results } };
 };
 
+const SearchFailedMessage = () => (
+    <div className="c-alert c-alert--warning">
+        <div className="c-alert__content">
+            Course search is currently unavailable. Please try again later, or{" "}
+            <a href="https://www.york.ac.uk/it-support/">contact IT Support</a>.
```
FergusMcGlynn, line 45: `{" "}` is something prettier added in automatically. This was originally a long-ish line, which prettier split into two. Because React strips whitespace at the start and end of lines, without an explicit space `{" "}`, the result of the above would be `Please try again later, orcontact IT Support`. It's a bit odd, but that's why it's there.

ncn1: Possibly a candidate for a prettier ignore comment?

ncn1 (approved):

src/pages/index.js
FergusMcGlynn: My feeling is to keep it how it is. I think we are going to have to get used to the fact that React strips whitespace, requiring us to artificially insert it on occasion (we've had to do it in the Proposals app too). While this particular line isn't _that_ long, there will be times when we have a very long line that needs to be split and then we'd need to do this, or something like it, then - I think we just need to get used to it.

src/pages/index.js
```diff
@@ -18,14 +19,32 @@ App.propTypes = {
...
+        isSuccessfulSearch = false;
+        searchResponseData = { results: [] };
+    }
 
-    return { props: { searchResults: data.results } };
+    return { props: { isSuccessfulSearch, searchResults: searchResponseData.results } };
 };
 
+const SearchFailedMessage = () => (
+    <div className="c-alert c-alert--warning">
```
ce666-york, line 42: I did a quick local merge of CFIN-123 into this branch to check these CSS classes, all looks good :+1:

src/pages/index.js
```diff
@@ -4,6 +4,7 @@ const App = (props) => {
     return (
         <>
             <h1>Course search results</h1>
+            {props.isSuccessfulSearch || <SearchFailedMessage />}
             {props.searchResults?.map((course) => (
                 <h2 key={course.liveUrl}>{course.title}</h2>
             ))}
```
ce666-york, line 10: I think I would prefer this to be an explicit `if`/`else` - it reads more nicely to me, and it prevents the confusing situation of seeing an error message along with a list of results (which could happen if the request logic changes later on).

Alternatively if you're expecting situations where there's a problem but results are still returned, then I'd advocate renaming it, as to me `isSuccessfulSearch` failing implies either no results or results you wouldn't want to show the user.

ce666-york (approved):

src/pages/index.js
FergusMcGlynn: I'm fine with the principle of an `if`/`else` type logic here - but it's not possible to have an `if`/`else` statement within a `return` statement. So I could potentially do something like
```
const App = (props) => {
    let errorMessage = "";
    if (!props.isSuccessfulSearch) {
        errorMessage = <SearchFailedMessage />
    }
   return (
      ...
      {errorMessage}
      ...
    }
```
But I'm not sure that's much cleaner. As for your second point - at present if `isSuccessfulSearch` is false then there'll be no results to return. What do you think?

FergusMcGlynn: Or maybe you were thinking something more like:
```
const App = (props) => {
    if (props.isSuccessfulSearch) {
        return (
            // happy path here
        );
    } else {
        return (
            <>
                <h1>Course search results</h1>
                <SearchFailedMessage />
            </>
        );
    }
```

FergusMcGlynn: This looks OK when the only common code between the happy path and the error path is `<h1>Course search results</h1>` - but when we have the pattern library stuff, it's a little bit more complicated...

ce666-york: I think my preference is around making the cases exclusive, be it by `if`/`else` or a ternary, so that latest example looks good -  it's not a big deal like I say, more for readability and one unlikely edge case.

I've no issue with the name, just checking you weren't expecting the logic to change later on (as the way the code was written implied that may have been the case).

ce666-york: I see this isn't quite as simple as it initially seemed; I approved this pull request as it's not a big issue for me, so happy to wait and see how things progress.

PR #17, opened 2020-09-25
CFIN-289: Tidy up index.js 

FergusMcGlynn:
This is a bit of a speculative PR - I won't be offended if people prefer the code without these changes. It is purely a refactor - no additional features have been added. @ce666-york said that it might be nice for the code to make it clearer what the page does when there's a failed search vs when there's a successful search. This is my attempt to address that (plus a couple of other tidy ups).

### DIFF
--- a/.xo-config.json
+++ b/.xo-config.json
@@ -12,6 +12,7 @@
     "indent":"off",
     "import/no-unassigned-import":"off",
     "react/react-in-jsx-scope":"off",
+    "react/jsx-no-useless-fragment":"off",
     "unicorn/filename-case":"off",
     "unicorn/import-index":"off",
     "unicorn/prevent-abbreviations":"off"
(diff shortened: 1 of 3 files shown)


.xo-config.json
```diff
@@ -12,6 +12,7 @@
     "indent":"off",
     "import/no-unassigned-import":"off",
     "react/react-in-jsx-scope":"off",
+    "react/jsx-no-useless-fragment":"off",
```
FergusMcGlynn, line 15: I've disabled this rule because of (I think) a bug in `eslint-plugin-react` where a JSX fragment (the bit between `<>` and `</>`) contains a `.map()`. It think that there is only one child, and hence no need for a fragment, but the map will actually return many children. See https://github.com/yannickcr/eslint-plugin-react/issues/2584. Note that that issue ends with a "merge" but we are using the "fixed" version and this is still a problem.

ncn1:
I think this makes it much clearer what happens when the search is successful and when it's not.

ncn1 (approved):

ce666-york (approved):
I prefer this version of the code, it's slightly longer but more explicit

PR #19, opened 2020-09-29
Cfin 292

ce666-york:


src/components/CourseSearchResults.js
```diff
@@ -0,0 +1,34 @@
+import { Course } from "./Course";
+import PropTypes from "prop-types";
+import { COURSE_MODEL } from "../constants/CourseModel";
+import React from "react";
+
+const CourseSearchResults = ({ isSuccessfulSearch, searchResults }) => {
```
ce666-york, line 6: I decided to refactor this out into a separate component, largely because the code is getting more complex but also because I felt uncomfortable with the index page declaring a `c-listings` div and then using `c-listing` in the `Course` component - this created an implicit coupling between the component and the page it's used in. This coupling still exists but it's between two components now, creating more of a component-subcomponent relationship.

mrgibbonsuoy (approved):

RichardAllum (approved):

src/components/CourseSearchResults.js
ChristianBeddows: Makes sense to me - and its really easy to understand what and why this component is doing

ChristianBeddows (approved):

ncn1 (approved):

FergusMcGlynn (approved):

PR #20, opened 2020-09-29
CFIN-292 Apply the pattern library default layout to the index page

ce666-york:
This PR goes through the [pattern library layout guide](https://www.york.ac.uk/pattern-library/layout/index.html) and applies some standard tags, page title, favicon etc.

I've also updated the page to use the full [basic grid](https://www.york.ac.uk/pattern-library/layout/grid.html) which gives us the standard responsive layout, which we can then update as we add new components to the page.

### DIFF
--- a/src/components/PageHead.js
+++ b/src/components/PageHead.js
@@ -1,0 +1,26 @@
+import Head from "next/head";
+import React from "react";
+
+const PageHead = () => {
+    return (
+        <Head>
+            <meta httpEquiv="X-UA-Compatible" content="IE=edge" />
+            <meta name="author" content="University of York" />
+            <meta name="viewport" content="width=device-width, initial-scale=1" />
+            <meta name="description" content="University of York Course search" />
+
+            <title>Course search, University of York</title>
+
+            <link rel="shortcut icon" href="https://www.york.ac.uk/static/stable/img/favicon.ico" />
+            <link rel="icon" type="image/x-icon" href="https://www.york.ac.uk/static/stable/img/favicon.ico" />
+
+            <script src="https://www.york.ac.uk/static/stable/js/modernizr.min.js" />
+            <script src="https://www.york.ac.uk/static/stable/js/app.min.js" />
+
+            <script src="//use.typekit.net/dvj8rpp.js" />
+            <script language="application/javascript">Typekit.load();</script>
+        </Head>
+    );
+};
+
+export { PageHead };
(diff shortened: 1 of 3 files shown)


RichardAllum (approved):

src/components/PageHead.js
```diff
@@ -0,0 +1,26 @@
+import Head from "next/head";
+import React from "react";
+
+const PageHead = () => {
```
RichardAllum, line 4: It might be worth adding a parameter for the course search that can go into the title, so when showing course results the title can be 'History Search Results - University of York' or similar. Not sure if it's necessary yet though.

ce666-york: That's a nice idea, it could be a way for users to more easily keep track of which searches they've run if they use multiple tabs, as we saw in the moderated UX research interviews.

ChristianBeddows (approved):

ncn1 (approved):

FergusMcGlynn (approved):

src/components/PageHead.js
```diff
@@ -0,0 +1,26 @@
...
+            <title>Course search, University of York</title>
+
+            <link rel="shortcut icon" href="https://www.york.ac.uk/static/stable/img/favicon.ico" />
+            <link rel="icon" type="image/x-icon" href="https://www.york.ac.uk/static/stable/img/favicon.ico" />
+
+            <script src="https://www.york.ac.uk/static/stable/js/modernizr.min.js" />
+            <script src="https://www.york.ac.uk/static/stable/js/app.min.js" />
+
+            <script src="//use.typekit.net/dvj8rpp.js" />
+            <script language="application/javascript">Typekit.load();</script>
```
FergusMcGlynn, line 21: I wonder if there's any use in testing these changes? If the user is going to experience changed behaviour as a result of this then it might be worth adding a test. Verifying that the page title is present and correct could be worth adding?

ce666-york: I agree, I did spend a short amount of time attempting to write a passing test without success so decided not to spend too much longer on it, as it didn't feel critical. After doing some more research it appears that Next components render [outside of the React tree](https://github.com/vercel/next.js/issues/8384), and so aren't picked up by React Testing Library.

[This issue](https://github.com/vercel/next.js/discussions/11060) suggests mocking `next/head` which I'm not particularly keen on, as it feels rather artificial and not testing the code that actually runs. Someone else comments on that page to say they'll probably use Cypress to test things like this.

Someone on [this StackOverflow question](https://stackoverflow.com/questions/62277514) wonders whether it's a good idea to use React Testing Library with Next at all. Seems to be working well enough for us for now, but perhaps worth bearing in mind.

I think the next step is to investigate Cypress at some point (which was already on our radar for the PMC).

PR #21, opened 2020-09-30
CFIN-285 add workflow to deploy app

RichardAllum:
The deployed app from our tests using CFIN-285 can be found here: https://9v5se9c1ce.execute-api.eu-west-1.amazonaws.com/v1

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -34,8 +34,7 @@ jobs:
 
       - name: Deploy - production
         if: github.ref == 'refs/heads/main'
-        run: |
-          npm run deploy
+        run: npm run deploy
         env:
           AWS_ACCESS_KEY_ID: ${{ secrets.PRODUCTION_AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.PRODUCTION_AWS_SECRET_ACCESS_KEY }}
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,43 @@
+name: Deploy to AWS
+
+on:
+  push:
+    branches:
+      - dev
+      - main
+  workflow_dispatch:
+
+jobs:
+  deploy_to_aws:
+    runs-on: ubuntu-latest
+    steps:
+
+      - name: Check out the application's source code
+        uses: actions/checkout@v2
+
+      - name: Install node
+        uses: actions/setup-node@v1
+        with:
+          node-version: 12
+
+      - name: Install application dependencies
+        run: npm ci
+
+      - name: Deploy - dev
+        if: github.ref == 'refs/heads/dev'
+        run: npm run deploy
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.DEV_AWS_ACCESS_KEY_ID }}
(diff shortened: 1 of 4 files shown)


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,43 @@
...
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.DEV_AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.DEV_AWS_SECRET_ACCESS_KEY }}
+          AWS_ACCOUNT_ID: ${{ secrets.DEV_AWS_ACCOUNT_ID }}
+          DEPLOY_ENVIRONMENT: dev
+
+      - name: Deploy - production
+        if: github.ref == 'refs/heads/main'
+        run: |
+          npm run deploy
```
ce666-york, line 38: This pipe might be unnecessary?

```suggestion
        run: npm run deploy
```

serverless.yml
```diff
@@ -5,6 +5,8 @@ provider:
   runtime: nodejs12.x
   stage: v1
   region: eu-west-1
+  # This role is required to deploy an application to AWS with an ESG Account. See https://wiki.york.ac.uk/display/AWS/AWS%3A+Github+Actions for details.
+  cfnRole: arn:aws:iam::${env:AWS_ACCOUNT_ID}:role/GithubActionsDeploymentRole
```
ce666-york, line 9: This doesn't look like the custom role you said you had to set up in dev, so I'm wondering how that's involved in the process?

ce666-york (approved):
Looks good :rocket: 

I suspect we may need to make more changes as we merge things in, as deploying a version of the app that uses the pattern library to my sandbox doesn't  seem to work due to 403's when requesting assets through the API gateway.

.github/workflows/deploy.yml
RichardAllum: It is, it was added when running a test task and setting debug mode before deploying.

serverless.yml
RichardAllum: That's a good question, the new account is a member of the GithubActionsDeployment Group but I'm not 100% on how necessary this is. The wiki recommends it however.

ce666-york: My understanding was that we were creating a new role with more permissions and attaching it to the existing deployment account so that we could then specify it here - is that not the case, and we're instead using tokens from a different account to run the deployments?

RichardAllum: I've had another look and it seems like we've been using this statement wrong. The role here refers to a role created per environment and not to a deployment user - users and roles are different in IAM. It still seems to be working here with an invalid cfnRole. We're currently using tokens from a user with more permissions than the default deployment account.

ce666-york: I don't think I have a full understanding of users/groups/roles/policies and how they interact so I can't comment further without doing some research. Essentially I think we should fully understand why we're doing what we're doing, and document it where it might be useful, especially anything specific to this project.

ChristianBeddows (approved):

serverless.yml
RichardAllum: After some work today we have discovered that  the GitHubActionsDeploymentRole acts  as a guide to which permissions cloudformation is able to use rather than a set of permissions used by a deployment, so we will continue to use it for now.

ncn1 (approved):

FergusMcGlynn (approved):

PR #22, opened 2020-10-01
CFIN 275 ensure accessibility

mrgibbonsuoy:
Addition of eslint-plugin-jsx-a11y to help with accessibility issues.
The test of this is to add in the <p> tags of src/components/course.js the following failing line: <iframe src="https://login.aws.york.ac.uk/" frameBorder="0">Hello</iframe>


FergusMcGlynn (changes requested):
This pull request seems to be adding in a bunch of commits that aren't relevant to this particular ticket. Perhaps the problem is that it has been set up to merge in to `dev` rather than into `CFIN-123`? I'd suggest deleting this PR and creating a new one that merges into `CFIN-123`.

mrgibbonsuoy:
> This pull request seems to be adding in a bunch of commits that aren't relevant to this particular ticket. Perhaps the problem is that it has been set up to merge in to `dev` rather than into `CFIN-123`? I'd suggest deleting this PR and creating a new one that merges into `CFIN-123`.

Hello Ferg: I'll resolve that! Thanks

PR #24, opened 2020-10-01
CFIN-275 Addition of eslint-plugin-jsx-a11y

mrgibbonsuoy:
Attempt 2. Apologies attempting a merge to the wrong branch previously!


package.json
```diff
@@ -30,8 +30,10 @@
         "babel-eslint": "^10.1.0",
         "babel-jest": "^26.3.0",
         "dotenv": "^8.2.0",
+        "eslint": "^7.10.0",
```
FergusMcGlynn, line 33: Do we need to include `eslint` explicitly as a dependency? I believe that `XO` will include it for us as it uses `eslint` under the bonnet. I'm just a little concerned that if we include `eslint` explicitly then that may lead to version conflicts with `XO` down the line. But if we do need `eslint` there then I'm OK with that.

mrgibbonsuoy: The install instructions say you need to install eslint first. 
From the readme https://github.com/jsx-eslint/eslint-plugin-jsx-a11y

Installation
If you are installing this plugin via eslint-config-airbnb, please follow these instructions.

You'll first need to install ESLint:

# npm
npm install eslint --save-dev

Next, install eslint-plugin-jsx-a11y:
# npm
npm install eslint-plugin-jsx-a11y --save-dev

FergusMcGlynn: My feeling is that these instructions are directed a folks who haven't already got `eslint` installed, either explicitly, or as part of another linting solution like `XO`. Could we try removing the `eslint` dependency, run `lint` and see whether it still works? If so, I'd be inclined to remove the `eslint` dependency.

mrgibbonsuoy: I'll branch form 123 and attempt it with just eslint-plugin-jsx-a11y and see how that goes.

mrgibbonsuoy:
I'll branch form 123 and attempt it with just eslint-plugin-jsx-a11y and see how that goes.

package.json
FergusMcGlynn: Thanks! It's fine to keep this PR open, BTW. You probably don't need to create a new branch - you could just experiment in your local git repo, edit `package.json` to remove the dependency, run `npm install`, try running `lint` and if all goes well then you can just do a commit and push. The PR will automatically update.

ce666-york: Yes, by all means use the same PR in future! It makes it much easier to track the conversations that have happened and see all of the context in one place.

mrgibbonsuoy: Sorry: I just felt that it was quicker and cleaner to go from 123 again inview of making sure no eslint overhead remained.

PR #25, opened 2020-10-01
CFIN 275 accessibility plugin without eslint plugin

mrgibbonsuoy:
Following a comment this branch does not have eslint installed as part of the eslint-plugin-jsx-a11y install.

### DIFF
--- a/.xo-config.json
+++ b/.xo-config.json
@@ -1,5 +1,5 @@
 {
-  "extends":["xo-react","prettier","prettier/react"],
+  "extends":["xo-react","prettier","prettier/react","plugin:jsx-a11y/recommended"],
   "overrides":[
     {
       "files":["**/*.test.js","e2e/**","jest/*"],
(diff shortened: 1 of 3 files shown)


.xo-config.json
```diff
@@ -1,5 +1,5 @@
 {
-  "extends":["xo-react","prettier","prettier/react"],
+  "extends":["xo-react","prettier","prettier/react","plugin:jsx-a11y/recommended"],
```
ce666-york, line 2: Does this need the `plugins` line from the other PR?

```suggestion
  "plugins":["jsx-a11y"],
  "extends":["xo-react","prettier","prettier/react","plugin:jsx-a11y/recommended"],
```

mrgibbonsuoy: Without it, it errors, but I'll just look to see what can be done.

ce666-york: To clarify, the `plugins` line that was in the other PR seems to be missing from this one. I would expect that it needs it.

PR #27, opened 2020-10-01
CFIN-298: Check code formatting and linting before PR

FergusMcGlynn:


mrgibbonsuoy (approved):

.github/workflows/checks.yml
```diff
@@ -1,8 +1,6 @@
 name: Run XO and Prettier
 on:
-  push:
-    branches:
-      - dev
```
ce666-york, line 5: I'd probably say leave this in, and even add `main` to the mix - there's little downside and it gives us more reassurance, plus we can create nifty GitHub badges 😉

.github/workflows/checks.yml
```diff
@@ -1,8 +1,6 @@
...
+  pull_request:
```
ce666-york, line 4: I like jobs having the option of being triggered manually, again little downside and you never know when you might need it

```suggestion
  workflow_dispatch:
```

RichardAllum (approved):

ncn1 (approved):

PR #28, opened 2020-10-02
CFIN-276 Retrieve assets from the correct URLs when running in API Gateway

ce666-york:
This is a small change to make Next account for the Stage in the AWS API Gateway URL when retrieving assets. I followed [these instructions in the Next documentation](https://nextjs.org/docs/api-reference/next.config.js/cdn-support-with-asset-prefix).

We might be able to remove this once we configure a stable domain name.

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -1,0 +1,6 @@
+const isProd = process.env.NODE_ENV === 'production'
+
+module.exports = {
+  // Use the CDN in production and localhost for development.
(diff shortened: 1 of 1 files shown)


FergusMcGlynn (approved):
We should be able to remove this once we have a domain name. In fact, we will probably need to remove it once we have a domain name - because it'll be looking for a `v1` path that won't exist.

ncn1 (approved):

PR #29, opened 2020-10-02
CFIN-299: Amend 'check' command to check formatting rather than fixing it

FergusMcGlynn:
Also adds a command `formatandcheck` that replaces the previous behaviour of `check`.


ncn1 (approved):
Thanks for adding the formatandcheck script as well!

ce666-york (approved):

PR #30, opened 2020-10-02
CFIN-123 As a prospective student I want to view a list of undergraduate course titles

ce666-york:
This is merging the whole `CFIN-123` story branch into `dev`, so the changes have been approved in previous PRs.

I have made one additional change here, which is altering `.prettierignore` to ignore `.next`, `.serverless` and a few more directories we don't want to run code formatting on (`.next` in particular was slowing it down).


package.json
```diff
@@ -7,13 +7,15 @@
     "build": "next build",
     "start": "next start",
     "deploy": "next build && sls deploy",
+    "undeploy": "sls undeploy",
```
FergusMcGlynn, line 10: I'm not sure when this line was introduced, but I don't think `sls undeploy` will work. It needs to be `sls remove`.
```suggestion
    "undeploy": "sls remove",
```

FergusMcGlynn (changes requested):

FergusMcGlynn (commented):
It looks like the `prettier`/`XO` automated check is failing - which should prevent the merge into `dev` from going ahead. Might need to run `check` to see what the problem is?

package.json
ce666-york: Oh I forgot about this! I've tried to use `npm run undeploy` a few times, which exists in the Courses API, so quickly added it here without thinking it through properly.

ce666-york:
The checks are failing because we now import the React pattern library components as part of `npm ci`, and so need a `.npmrc` file. I can look into adding this into the workflow.

FergusMcGlynn (approved):

PR #33, opened 2020-10-06
CFIN-312 Update workflow and severless to use custom domain/URL

ChristianBeddows:
Spoke to Chris earlier and for now we're going to use courses.app.york.ac.uk. This is to prevent people finding the prototype and getting confused by its lack of functionality before first release.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -44,6 +44,8 @@ jobs:
           AWS_SECRET_ACCESS_KEY: ${{ secrets.DEV_AWS_SECRET_ACCESS_KEY }}
           AWS_ACCOUNT_ID: ${{ secrets.DEV_AWS_ACCOUNT_ID }}
           DEPLOY_ENVIRONMENT: dev
+          SSL_CERTIFICATE_ARN: ${{ secrets.DEV_SSL_CERTIFICATE_ARN }}
+          DOMAIN_NAME: courses.dev.app.york.ac.uk
 
       - name: Deploy - production
         if: github.ref == 'refs/heads/main'
@@ -53,3 +55,5 @@ jobs:
           AWS_SECRET_ACCESS_KEY: ${{ secrets.PRODUCTION_AWS_SECRET_ACCESS_KEY }}
           AWS_ACCOUNT_ID: ${{ secrets.PRODUCTION_AWS_ACCOUNT_ID }}
           DEPLOY_ENVIRONMENT: prod
+          SSL_CERTIFICATE_ARN: ${{ secrets.PRODUCTION_SSL_CERTIFICATE_ARN }}
+          DOMAIN_NAME: courses.app.york.ac.uk
--- a/next.config.js
+++ b/next.config.js
@@ -1,6 +1,0 @@
-const isProd = process.env.NODE_ENV === "production";
-
-module.exports = {
-    // Retrieve assets from the correct URL that accounts for the Stage when running in AWS API Gateway.
-    assetPrefix: isProd ? "/v1" : "",
-};
--- a/serverless.yml
+++ b/serverless.yml
@@ -75,15 +75,13 @@ package:
     - node_modules/micromatch/**
 
 resources:
-
   Conditions:
     UseCustomDomainName: !Not
       - !Equals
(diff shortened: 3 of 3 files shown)


ce666-york (changes requested):
This all looks good so far! However I think you still need to tell the app to use the domain name - at the moment you're declaring custom variables in `serverless.yml`, but they're not being used anywhere yet.

In the Courses API this is handled by the `ApiGatewayCustomDomainName`, `BasePathMapping` and `AppCNAME` custom resources, which I think we'll need to replicate. One thing we might be able to do is just declare an `AppCNAME` resource, and rely on the `serverless-domain-manager` plugin for the rest, as detailed in this post https://www.serverless.com/blog/serverless-api-gateway-domain (I think we'd set `createRoute53Record` to `false` and use the `certificateName` property, just from a skim of that, but I could be wrong).

It might be a good idea to configure your local machine to deploy to the dev environment so you can test things out, and not have to push and run a workflow each time.

FergusMcGlynn:
> This all looks good so far! However I think you still need to tell the app to use the domain name - at the moment you're declaring custom variables in `serverless.yml`, but they're not being used anywhere yet.
> 
> In the Courses API this is handled by the `ApiGatewayCustomDomainName`, `BasePathMapping` and `AppCNAME` custom resources, which I think we'll need to replicate. One thing we might be able to do is just declare an `AppCNAME` resource, and rely on the `serverless-domain-manager` plugin for the rest, as detailed in this post https://www.serverless.com/blog/serverless-api-gateway-domain (I think we'd set `createRoute53Record` to `false` and use the `certificateName` property, just from a skim of that, but I could be wrong).
> 
> It might be a good idea to configure your local machine to deploy to the dev environment so you can test things out, and not have to push and run a workflow each time.

I have struggled with `serverless-domain-manager` - and I've submitted an issue in their Github repo because of it. `serverless-domain-manager` will set up a CloudFront distribution for you behind-the-scenes, but there's no way (as far as I can tell) of knowing what that CloudFront distribution's URL is within `serverless.yml` - and therefore no way of setting up a CNAME from your custom domain name to that CloudFront URL.

The solution I came up with wasn't too much hassle - you can [view it here](https://github.com/FergusMcGlynn/nextjs-demo/blob/master/serverless.yml). Ignore the stuff about generating the SSL certificate, as that won't be relevant here.

FergusMcGlynn:
BTW if you have set up the certificate in `eu-west-1` then the solution that I linked to should work as is. If you set up the certificate in `us-east-1` then you'd need to make a couple of modifications. Rather than

```
    ApiGatewayCustomDomainName:
      Type: AWS::ApiGateway::DomainName
      Properties:
        RegionalCertificateArn: !Ref IssuedSSLCertificate
        DomainName: ${self:custom.domainName}
        EndpointConfiguration:
          Types:
            - REGIONAL
```
it would be
```
    ApiGatewayCustomDomainName:
      Type: AWS::ApiGateway::DomainName
      Properties:
        CertificateArn: !Ref IssuedSSLCertificate
        DomainName: ${self:custom.domainName}
        EndpointConfiguration:
          Types:
            - EDGE
```
and rather than
```
    AppCNAME:
      Type: Custom::CNAME
      Properties:
        ServiceToken: "arn:aws:sns:eu-west-1:230504789214:RequestRecordSet"
        Source: ${self:custom.domainName}.
        Target: !Join [ '', [ !GetAtt ApiGatewayCustomDomainName.RegionalDomainName, '.' ] ]
```
it would be
```
    AppCNAME:
      Type: Custom::CNAME
      Properties:
        ServiceToken: "arn:aws:sns:eu-west-1:230504789214:RequestRecordSet"
        Source: ${self:custom.domainName}.
        Target: !Join [ '', [ !GetAtt ApiGatewayCustomDomainName.DistributionDomainName, '.' ] ]
```

For various reasons, I think it might be better to define the certificate in `eu-west-1` rather than `us-east-1` for the time being - until we want to use CloudFront, or until we decide to go with `courses.york.ac.uk`. In which case, ignore the above changes that I've outlined.

FergusMcGlynn (approved):
Looking good!

ce666-york (approved):

PR #37, opened 2020-10-07
CFIN-307: Create re-usable tag component

FergusMcGlynn:
This PR adds components `<TagList>` and `<Tag>`. The tag includes an `icon` parameter which aligns with the latest designs. These components could be used like 
```
<TagList>
    <Tag icon="clock" mainText="Start September 2021" />
    <Tag icon="graduation-cap" mainText="Undergraduate />
</TagList>
```
although I imagine that we'll want to [specialise](https://reactjs.org/docs/composition-vs-inheritance.html#specialization) the `<Tag>` component and create `<AwardTag>` and `<StartDateTag>` so that ultimately we use tags more like
```
<TagList>
    <StartDateTag year="2021" />
    <AwardTag award="BSc" />
</TagList>
```
and `<AwardTag>`, for example, will make use of the generic `<Tag>`:
```
const AwardTag ({ award }) => <Tag icon="institution" mainText={award} />
```
The `<Tag>` component will probably need to be modified down the line to accommodate secondary text (e.g `single subject`) depending on the outcome of the design process. Also, the proposed design for the "year in industry"/"year abroad" tag will probably require a bespoke component as it has a different structure to all the other tags.

### DIFF
--- a/.xo-config.json
+++ b/.xo-config.json
@@ -14,6 +14,9 @@
     "import/no-unassigned-import": "off",
     "react/react-in-jsx-scope": "off",
     "react/jsx-no-useless-fragment": "off",
+    "react/self-closing-comp": ["error", {
+      "html": false
+    }],
     "unicorn/filename-case": "off",
     "unicorn/import-index": "off",
     "unicorn/prevent-abbreviations": "off"
--- a/.xo-config.json
+++ b/.xo-config.json
@@ -14,12 +14,6 @@
     "import/no-unassigned-import": "off",
     "react/react-in-jsx-scope": "off",
     "react/jsx-no-useless-fragment": "off",
-    "react/self-closing-comp": [
-      "error",
-      {
-        "html": false
-      }
-    ],
     "unicorn/filename-case": "off",
     "unicorn/import-index": "off",
     "unicorn/prevent-abbreviations": "off"
--- a/.xo-config.json
+++ b/.xo-config.json
@@ -14,9 +14,12 @@
     "import/no-unassigned-import": "off",
     "react/react-in-jsx-scope": "off",
     "react/jsx-no-useless-fragment": "off",
-    "react/self-closing-comp": ["error", {
-      "html": false
-    }],
+    "react/self-closing-comp": [
+      "error",
+      {
+        "html": false
+      }
+    ],
     "unicorn/filename-case": "off",
     "unicorn/import-index": "off",
     "unicorn/prevent-abbreviations": "off"
--- a/src/components/Tag.js
+++ b/src/components/Tag.js
@@ -6,7 +6,7 @@ const Tag = ({ icon, mainText }) => (
         <span className="c-tag">
             <i aria-hidden className={`c-tag__icon c-icon c-icon--${icon}`} data-testid="tag-icon" />
             {mainText}
-        </span>
+        </span>{" "}
     </li>
 );
 
--- a/src/components/Tag.js
+++ b/src/components/Tag.js
@@ -6,7 +6,7 @@ const Tag = ({ icon, mainText }) => (
         <span className="c-tag">
             <i aria-hidden className={`c-tag__icon c-icon c-icon--${icon}`} data-testid="tag-icon" />
             {mainText}
-        </span>{" "}
+        </span>
     </li>
 );
 
--- a/src/components/Tag.js
+++ b/src/components/Tag.js
@@ -6,7 +6,7 @@ const Tag = ({ icon, mainText }) => (
         <span className="c-tag">
             <i aria-hidden className={`c-tag__icon c-icon c-icon--${icon}`} data-testid="tag-icon"></i>
             {mainText}
-        </span>
+        </span>{" "}
     </li>
 );
 
--- a/src/components/Tag.js
+++ b/src/components/Tag.js
@@ -1,0 +1,18 @@
+import PropTypes from "prop-types";
+import React from "react";
+
+const Tag = ({ icon, mainText }) => (
+    <li className="c-tag-list__item">
+        <span className="c-tag">
+            <i aria-hidden className={`c-tag__icon c-icon c-icon--${icon}`} data-testid="tag-icon"></i>
+            {mainText}
+        </span>
+    </li>
+);
+
+Tag.propTypes = {
+    icon: PropTypes.string.isRequired,
+    mainText: PropTypes.string.isRequired,
+};
+
+export { Tag };
--- a/src/components/Tag.js
+++ b/src/components/Tag.js
@@ -11,7 +11,7 @@ const Tag = ({ icon, mainText }) => (
 );
 
 Tag.propTypes = {
-    icon: PropTypes.string.isRequired,
+    icon: PropTypes.string,
     mainText: PropTypes.string.isRequired,
 };
 
--- a/src/components/Tag.js
+++ b/src/components/Tag.js
@@ -4,7 +4,7 @@ import React from "react";
 const Tag = ({ icon, mainText }) => (
     <li className="c-tag-list__item">
         <span className="c-tag">
-            <i aria-hidden className={`c-tag__icon c-icon c-icon--${icon}`} data-testid="tag-icon"></i>
+            <i aria-hidden className={`c-tag__icon c-icon c-icon--${icon}`} data-testid="tag-icon" />
             {mainText}
         </span>{" "}
     </li>
--- a/src/components/TagList.js
+++ b/src/components/TagList.js
@@ -1,17 +1,19 @@
 import React from "react";
 
-const TagList = ({ tags }) => {
+const TagList = ({ items }) => {
     return (
         <ul>
-            {tags.map((item) => {
-                return (
-                    <li key={item} aria-labelledby={`${item}-tag`}>
-                        <span id={`${item}-tag`}>{item}</span>
-                    </li>
-                );
-            })}
+            {items.map((item) => (
+                <Tag key={item} name={item} />
+            ))}
         </ul>
     );
 };
+
+const Tag = ({ name }) => (
+    <li aria-labelledby={`${name}-tag`}>
+        <span id={`${name}-tag`}>{name}</span>
+    </li>
+);
 
 export { TagList };
--- a/src/components/TagList.js
+++ b/src/components/TagList.js
@@ -1,30 +1,12 @@
 import React from "react";
 import PropTypes from "prop-types";
 
-const TagList = ({ items }) => {
-    return (
-        <ul className="c-tag-list">
-            {items.map((item) => (
-                <Tag key={item} name={item} />
-            ))}
-        </ul>
-    );
+const TagList = (props) => {
+    return <ul className="c-tag-list">{props.children}</ul>;
 };
 
 TagList.propTypes = {
-    items: PropTypes.arrayOf(PropTypes.string),
-};
-
-const Tag = ({ name }) => (
-    <li className="c-tag-list__item" aria-labelledby={`${name}-tag`}>
-        <span className="c-tag" id={`${name}-tag`}>
-            {name}
-        </span>
-    </li>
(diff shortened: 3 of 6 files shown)


.xo-config.json
```diff
@@ -14,6 +14,9 @@
     "import/no-unassigned-import": "off",
     "react/react-in-jsx-scope": "off",
     "react/jsx-no-useless-fragment": "off",
+    "react/self-closing-comp": ["error", {
+      "html": false
+    }],
```
FergusMcGlynn, line 19: This is to deal with the `<i>` tag used for icons NOT self closing. [The closing `</i>` is mandatory](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/i), but ESLint complains that because there's no body (when used for an icon) it should self close.

src/tests/components/Tag.test.js
```diff
@@ -0,0 +1,13 @@
...
+import { Tag } from "../../components/Tag";
+import React from "react";
+
+describe("Tag", () => {
+    it("displays tag containing given icon and text", () => {
+        render(<Tag icon="battery-full" mainText="Battery is full" />);
+
+        expect(screen.getByRole("listitem")).toHaveTextContent("Battery is full");
+        expect(screen.getByRole("listitem")).toContainElement(screen.getByTestId("tag-icon"));
+        expect(screen.getByTestId("tag-icon")).toHaveClass("c-icon--battery-full");
```
FergusMcGlynn, line 11: I don't love the idea of testing for a `class` attribute - but there isn't really any other way of verifying that the correct icon is being displayed. The icon needs to be [`aria-hidden` so that it doesn't confuse a screen reader](https://www.york.ac.uk/pattern-library/css-components/icons.html), and that limits testing options somewhat.

.xo-config.json
```diff
@@ -14,6 +14,12 @@
     "import/no-unassigned-import": "off",
     "react/react-in-jsx-scope": "off",
     "react/jsx-no-useless-fragment": "off",
+    "react/self-closing-comp": [
+      "error",
+      {
+        "html": false
+      }
+    ],
```
FergusMcGlynn, line 22: This is to deal with the `<i>` tag used for icons NOT self closing. [The closing `</i>` is mandatory](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/i), but ESLint complains that because there's no body (when used for an icon) it should self close.

mrgibbonsuoy (approved):

src/tests/components/Tag.test.js
FergusMcGlynn: I did play around with setting accessible names for the tags, e.g.
```<li aria-label="foo bar">foo bar</li>``` which would lead to a test like ```expect(screen.getByRole("listitem", {name: "foo bar"})).toHaveTextContent("foo bar");``` which I don't find particularly intuitive. I also tried
```<li aria-labelledby="some-value"><span id="some-value">foo bar</span></li>``` which would lead to a test like ```expect(screen.getByRole("listitem", {name: "foo bar"})).toBeInTheDocument();```, which is a bit more intuitive, but then every `<span>` would need a unique `id`, and the only unique item of data between courses we have is the course URL, and it didn't feel very clean passing in the course URL to each tag simply so that it could produce a unique `id` for each `<span>`. Also, I came to the conclusion that these did nothing to improve accessibility, and a screen reader is perfectly happy reading out a list without and `aria-label` assistance in this case.

File: src/tests/components/Tag.test.js [hunk 4]
```diff
@@ -0,0 +1,13 @@
...
+import { Tag } from "../../components/Tag";
+import React from "react";
+
+describe("Tag", () => {
+    it("displays tag containing given icon and text", () => {
+        render(<Tag icon="battery-full" mainText="Battery is full" />);
+
+        const tag = screen.getByRole("listitem");
+        expect(tag).toHaveTextContent("Battery is full");
+        expect(within(tag).getByTestId("tag-icon")).toHaveClass("c-icon--battery-full");
```
FergusMcGlynn, line 11: Re-adding this comment as it was removed when I tidied up this code a little...

I don't love the idea of testing `byTestId` or by checking a `class` attribute - but there isn't really any other way of verifying that the correct icon is being displayed. The icon needs to be `aria-hidden` so that it doesn't confuse a screen reader, and that limits testing options somewhat.

 
I did play around with setting accessible names for the tags, e.g.
```<li aria-label="foo bar">foo bar</li>``` 
which would lead to a test like 
```expect(screen.getByRole("listitem", {name: "foo bar"})).toHaveTextContent("foo bar");``` 
which I don't find particularly intuitive. I also tried
```<li aria-labelledby="some-value"><span id="some-value">foo bar</span></li>``` 
which would lead to a test like 
```expect(screen.getByRole("listitem", {name: "foo bar"})).toBeInTheDocument();``` 
which is a bit more intuitive, but then every `<span>` would need a unique `id`, and the only unique item of data between courses we have is the course URL, and it didn't feel very clean passing in the course URL to each tag simply so that it could produce a unique `id` for each `<span>`. I came to the conclusion that these changes did nothing to improve accessibility (a screen reader is perfectly happy reading out a list without any `aria-label` assistance in this case), nor test readability.

**Comment by FergusMcGlynn** [comment]
This is how the tags look when added to a course search result

image

File: src/components/Tag.js [hunk 5]
```diff
@@ -0,0 +1,18 @@
+import PropTypes from "prop-types";
+import React from "react";
+
+const Tag = ({ icon, mainText }) => (
+    <li className="c-tag-list__item">
+        <span className="c-tag">
+            <i aria-hidden className={`c-tag__icon c-icon c-icon--${icon}`} data-testid="tag-icon"></i>
+            {mainText}
+        </span>{" "}
```
FergusMcGlynn, line 9: The explicit whitespace here (`{" "}`) is to work around React's stripping of all whitespace, which results in tags butting right up against each other.

src/tests/components/Tag.test.js
ncn1: These sound like good reasons to use it - I agree should only be used if there isn't a better alternative.

ncn1 (approved):

mrgibbonsuoy (approved):

src/components/Tag.js
RichardAllum: In the style guide the tags are defined as links with an href of '#' which seems to give them padding, would that solve the issue of them sitting next to each other?

src/components/Tag.js
```diff
@@ -0,0 +1,18 @@
...
+    </li>
+);
+
+Tag.propTypes = {
+    icon: PropTypes.string.isRequired,
```
RichardAllum, line 14: Should the icon be required? We want it for the V2 designs, which is a point in favour, but having it not required would make it easier to test other designs.

.xo-config.json
ChristianBeddows: If you have a tag such as `<i src="whatever.com" />` does that still throw the error?

src/components/Tag.js
FergusMcGlynn: Unfortunately not - I've tried that locally and the tags still butt up against each other. Also I'm not sure I'd want these to be links just yet, even if the `href` was just `#`.

ChristianBeddows:
I really like the style of the Tag components, nice work!

src/components/TagList.js
```diff
@@ -0,0 +1,12 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+const TagList = (props) => {
+    return <ul className="c-tag-list">{props.children}</ul>;
+};
+
+TagList.propTypes = {
+    children: PropTypes.node,
```
RichardAllum, line 9:
```suggestion
    children: PropTypes.node.isRequired,
```
Do we expect search results without any tags? (We may do now until the other tasks are finished, which is why I'm asking)

src/components/Tag.js
ChristianBeddows: Using whitespace as a method of styling the page seems a bit smelly. Also, does `&nbsp;` make more sense than this string?

ChristianBeddows (approved):

.xo-config.json
FergusMcGlynn: ESLint wouldn't complain, because it has an empty body and is self-closing, which is what it likes. However, it would not be legal HTML. I've tried it as a self-closing tag (i.e. without the closing `</i>`) and the styling gets stuffed up a bit, possibly because it's not legal and browser doesn't render it properly (although I'm not sure).

src/components/Tag.js
RichardAllum: I had a look at using &nbsp; and it looks like React also strips it from between tags, so the way to add a space is via `{" "}`

RichardAllum (approved):

src/components/Tag.js
FergusMcGlynn: I agree that it's a bit smelly. I don't think `&nbsp;` is any better really - it just another explicit space character. We could take out the space here and just leave the tags butting up against each other - and lobby for change in the pattern library so that this doesn't happen maybe?

.xo-config.json
RichardAllum: I think the legality of it is a purely technical issue and browsers have no trouble with rendering self-closing `<i>` tags, the Pattern Library components use a self-closing tag and show no issues. That said there's nothing wrong with choosing the technically correct option.

src/tests/components/Tag.test.js
RichardAllum: https://www.w3.org/WAI/GL/wiki/Providing_a_Semantically_Identified_Icon_Font_with_role%3Dimg W3 recommend that the "img" role is used for icons for identifying them, would that be semantically better? I think "tag-icon" makes a lot of sense in the tests though.

.xo-config.json
FergusMcGlynn: I've just tried it again using a self closing tag and it seems to behave fine. When I try it with two tags, one `<i>` self-closing and the other explicitly closing, they do get presented differently, but if they're both self-closing then it's fine. I will revert to self-closing and remove this ESLint exception.

RichardAllum (approved):

src/components/Tag.js
FergusMcGlynn: Fair point - I will make it not required

src/components/Tag.js
ncn1: Does sound like we need some css padding around these - either in the pattern library - or we do it locally? However, we are going to be using some new tag definitions eventually to match our preferred design - maybe that's the time to address it.

src/tests/components/Tag.test.js
FergusMcGlynn: The `img` role would only have any use if the icons aren't `aria-hidden`. I think it makes sense for them to be `aria-hidden` as they're purely presentational. It's also [recommended by the pattern library documentation](https://www.york.ac.uk/pattern-library/css-components/icons.html).

src/components/Tag.js
FergusMcGlynn: That's a fair point. I will remove the explicit whitespace, and we can address the lack of padding either in local css or, ideally, in the pattern library.

src/components/TagList.js
FergusMcGlynn: Yes, that seems like a reasonable assumption

FergusMcGlynn:
Here's how it look with the explicit whitespace removed and the tags butting up against each other. I'm fine with this for the time being. As Nicola says, we will almost certainly be re-designing these tags soon and that'll give us an opportunity to consider appropriate CSS for separating the tags.

image

ce666-york (approved):
I've caught up on all the discussions that have happened on this PR, all sounds sensible to me. 

I'd be tempted to leave tag spacing in for now and tidy it up when we update the tags - feels like it's such a small imperfection it's worth prioritising the current design over the code in this case. That's just my opinion though.

:+1:

PR #36, opened 2020-10-07
CFIN-303 allow users to enter a search term

ce666-york:
This change adds a search box and basic search functionality to Course Search. You can try it out in my AWS sandbox account at https://y3cdpz4cgk.execute-api.eu-west-1.amazonaws.com/v1.

If a search term hasn't been entered it uses a default one for now - there's a separate story [CFIN-253](https://jira.york.ac.uk/browse/CFIN-253) about seeing a sensible list of courses when not entering a search term.

The following will be covered in separate PRs:
- Adding a sensible message when no search results are returned.
- Accessibility checks

---

cfin-303

PR #38, opened 2020-10-07
CFIN-304 display a message when no search results are returned

ce666-york:
As with #36 you can try this out in my AWS sandbox at https://y3cdpz4cgk.execute-api.eu-west-1.amazonaws.com/v1.

cfin-304

### DIFF
--- a/src/components/CourseSearchResults.js
+++ b/src/components/CourseSearchResults.js
@@ -8,7 +8,7 @@ const CourseSearchResults = ({ isSuccessfulSearch, searchResults }) => {
         return <SearchFailedMessage />;
     }
 
-    if (!searchResults) {
+    if (!searchResults || !searchResults.length) {
         return <NoSearchResultsMessage />;
     }
 
--- a/src/components/CourseSearchResults.js
+++ b/src/components/CourseSearchResults.js
@@ -8,7 +8,7 @@ const CourseSearchResults = ({ isSuccessfulSearch, searchResults }) => {
         return <SearchFailedMessage />;
     }
 
-    if (!searchResults || !searchResults.length) {
+    if (!searchResults?.length) {
         return <NoSearchResultsMessage />;
     }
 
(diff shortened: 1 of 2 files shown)


ChristianBeddows:
I really like the sandbox preview for your code, a great way of visualising the changes.

ChristianBeddows (approved):

src/components/CourseSearchResults.js
```diff
@@ -8,6 +8,10 @@ const CourseSearchResults = ({ isSuccessfulSearch, searchResults }) => {
         return <SearchFailedMessage />;
     }
 
+    if (!searchResults?.length) {
```
FergusMcGlynn, line 11: Very minor point, but could this be simplified?
```suggestion
    if (!searchResults) {
```

FergusMcGlynn (approved):

src/components/CourseSearchResults.js
```diff
@@ -31,4 +35,10 @@ const SearchFailedMessage = () => (
     </div>
 );
 
+const NoSearchResultsMessage = () => (
+    <div className="c-alert c-alert--info">
```
RichardAllum, line 39: I think we could do this with a pattern Library `<Alert>` component with a className of info
```suggestion
    <Alert className="info">Sorry, your search did not return any results.</Alert>
```

ce666-york: I didn't clock this, yes this feels like an improvement (although an `InfoAlert` would be even better)! As we have another alert already being used elsewhere, I'll look at refactoring them both at the same time in a separate commit

src/components/CourseSearchResults.js
ce666-york: Sounded good to me so I committed the suggestion, but it caused the tests to fail!

`!searchResults` just checks for the existence of the array I guess, and not the content. I've decided to make it `!searchResults || !searchResults.length` to be a bit more robust

FergusMcGlynn: Oops... Sorry! :-/

RichardAllum (approved):

src/components/CourseSearchResults.js
ce666-york: No problem at all, it's a learning experience!

PR #39, opened 2020-10-07
CFIN-305 Improve course search page accessibility

ce666-york:
This change covers two accessibility issues flagged by WAVE, if anyone has strong opinions about this I'd welcome any discussion!

- **ERROR: Set a page language** - this is fairly self-explanatory, and is done in a special `pages/_document.js` file which Next understands. I've copied the default page template as shown in [the example in the documentation](https://nextjs.org/docs/advanced-features/custom-document) and just added the `lang` attribute (this specific example is given as a reason to use `_document.js`).

- **ALERT: Fieldset without a legend** - WAVE flagged up that every form fieldset should have a legend; or, if a legend is deemed unnecessary, you probably don't need a fieldset. I read through some of the [accessibility guidelines around creating understandable pages through labels or instructions](https://www.w3.org/WAI/WCAG21/quickref/#labels-or-instructions) (section 3.3.2) which provides a lot of examples of more complex forms that require validation, or information about what the outcome of submitting the form will be (e.g. [one](https://www.w3.org/WAI/WCAG21/Techniques/general/G184.html), [two](https://www.w3.org/WAI/WCAG21/Techniques/general/G89.html), [three](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA1.html) etc.). Given the simplicity of the search form, I decided that we probably don't need a legend and therefore don't need to group the form controls under a fieldset.

### DIFF
--- a/src/pages/_document.js
+++ b/src/pages/_document.js
@@ -1,0 +1,22 @@
+import Document, { Html, Head, Main, NextScript } from "next/document";
+
+class MyDocument extends Document {
+    static async getInitialProps(ctx) {
+        const initialProps = await Document.getInitialProps(ctx);
+        return { ...initialProps };
+    }
+
+    render() {
+        return (
+            <Html lang="en">
+                <Head />
+                <body>
+                    <Main />
+                    <NextScript />
+                </body>
+            </Html>
+        );
+    }
+}
+
+export default MyDocument;
--- a/src/components/Search.js
+++ b/src/components/Search.js
@@ -1,6 +1,5 @@
 import {
     FormElement,
-    FormFieldset,
     FormInputText,
     Grid,
     GridBoxFull,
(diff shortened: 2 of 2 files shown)


FergusMcGlynn (approved):

ncn1 (approved):

src/pages/_document.js
```diff
@@ -0,0 +1,22 @@
...
+class MyDocument extends Document {
+    static async getInitialProps(ctx) {
+        const initialProps = await Document.getInitialProps(ctx);
+        return { ...initialProps };
+    }
+
+    render() {
+        return (
+            <Html lang="en">
+                <Head />
```
RichardAllum, line 12: Is the reason we're using Next's `Head` rather than `PageHead` here that Next does some jiggery-pokery to add things into the Head itself?

RichardAllum (approved):

src/pages/_document.js
ce666-york: The documentation page I linked says `Head` from `next/document` and `Head` from `next/head` are two separate things, and used for different reasons. 

This one I think is more fundamental - I tried to refactor `PageHead` tags into this and it complained saying things like the `title` and various meta tags shouldn't be in it, so it feels like `next/head` (and therefore the `PageHead` component) is still the right place for it.

PR #44, opened 2020-10-09
CFIN-308 add start date tag

mrgibbonsuoy:
Addition of the start year.

### DIFF
--- a/src/components/CourseDetails.js
+++ b/src/components/CourseDetails.js
@@ -14,7 +14,8 @@ const CourseDetails = ({ course }) => {
 
 const StartDateTag = ({ yearOfEntry }) => {
     if (yearOfEntry) {
-        return <Tag icon="calendar" mainText={`Starts ${yearOfEntry}`} />;
+        const startYear = yearOfEntry.slice(0, 4);
+        return <Tag icon="calendar" mainText={`Starts ${startYear}`} />;
     }
 
     return null;
--- a/src/components/CourseDetails.js
+++ b/src/components/CourseDetails.js
@@ -2,13 +2,26 @@ import React from "react";
 import { COURSE_MODEL } from "../constants/CourseModel";
 import { TagList } from "./TagList";
 import { Tag } from "./Tag";
+import PropTypes from "prop-types";
 
 const CourseDetails = ({ course }) => {
     return (
         <TagList>
-            <Tag icon="calendar" mainText={`Starts ${course.yearOfEntry}`} />
+            <StartDateTag yearOfEntry={course.yearOfEntry} />
         </TagList>
     );
+};
+
+const StartDateTag = ({ yearOfEntry }) => {
+    if (yearOfEntry) {
+        return <Tag icon="calendar" mainText={`Starts ${yearOfEntry}`} />;
+    }
+
+    return null;
+};
+
+StartDateTag.propTypes = {
+    yearOfEntry: PropTypes.string,
 };
 
 CourseDetails.propTypes = {
--- a/src/components/CourseDetails.js
+++ b/src/components/CourseDetails.js
@@ -6,7 +6,7 @@ import { Tag } from "./Tag";
 const CourseDetails = ({ course }) => {
     return (
         <TagList>
-            <Tag icon="twitter" mainText="example" />
+            <Tag icon="calendar" mainText={`Starts ${course.yearOfEntry}`} />
         </TagList>
     );
 };
--- a/src/components/CourseDetails.js
+++ b/src/components/CourseDetails.js
@@ -1,0 +1,18 @@
+import React from "react";
+import { COURSE_MODEL } from "../constants/CourseModel";
+import { TagList } from "./TagList";
+import { Tag } from "./Tag";
+
+const CourseDetails = ({ course }) => {
+    return (
+        <TagList>
+            <Tag icon="twitter" mainText="example" />
+        </TagList>
+    );
+};
+
+CourseDetails.propTypes = {
+    course: COURSE_MODEL,
+};
+
+export { CourseDetails };
--- a/src/tests/components/CourseDetails.test.js
+++ b/src/tests/components/CourseDetails.test.js
@@ -10,10 +10,10 @@ describe("Course details", () => {
 
         render(<CourseDetails course={exampleCourse} />);
 
-        const tags = screen.getAllByRole("listitem");
+        const tag = screen.getByRole("listitem");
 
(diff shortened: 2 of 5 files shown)


src/tests/components/CourseDetails.test.js
```diff
@@ -0,0 +1,18 @@
+import { render, screen, within } from "@testing-library/react";
+import React from "react";
+import { CourseDetails } from "../../components/CourseDetails";
+
+describe("TagList", () => {
+    it("displays tags labelled with given inputs", () => {
```
RichardAllum, line 6:
```suggestion
    it("displays course metadata in tags with icons where appropriate", () => {
```

src/components/CourseDetails.js
```diff
@@ -0,0 +1,18 @@
+import React from "react";
+import { COURSE_MODEL } from "../constants/CourseModel";
+import { TagList } from "./TagList";
+import { Tag } from "./Tag";
+
+const CourseDetails = ({ course }) => {
+    return (
+        <TagList>
+            <Tag icon="calendar" mainText={`Starts ${course.yearOfEntry}`} />
```
RichardAllum, line 9: Do we want to not show these tags when they're not present in the results?
I think they probably will be in all cases for this, but might not be for every piece of metadata. We could apply this by doing something like
```suggestion
            { course.yearOfEntry && <Tag icon="calendar" mainText={`Starts ${course.yearOfEntry}`} />}
```
This is better for this component in the long run, however it will break TagList if we have a course with no metadata - TagList requires a list of nodes as children.

RichardAllum (approved):

src/tests/components/CourseDetails.test.js
```diff
@@ -0,0 +1,18 @@
+import { render, screen, within } from "@testing-library/react";
+import React from "react";
+import { CourseDetails } from "../../components/CourseDetails";
+
+describe("TagList", () => {
```
FergusMcGlynn, line 5: No tests for what happens when a course doesn't have a `yearOfEntry` field?

src/components/CourseDetails.js
FergusMcGlynn: I quite like Richard's suggestion. Another approach might be to specialise the `<Tag>` with something like
```
const StartDateTag = ({ yearOfEntry }) => {
    if (yearOfEntry) {
        return <Tag icon="calendar" mainText={`Starts ${course.yearOfEntry}`} />
    }
}
```
(that's probably not totally legal syntax but just a general idea) - and then call it with
```
<TagList>
    <StartDateTag yearOfEntry={course.yearOfEntry} />
</TagList>
```

FergusMcGlynn: Also there's no problem modifying `<TagList>` to NOT require children, if needed, for the time being

FergusMcGlynn, line 9: In the acceptance criteria for this ticket, it says `The year of entry of each course is visible on the search results page`. If the data coming back from the Courses API for `yearOfEntry` formatted like `2021/22` then I think we ought to apply some re-formatting so that it's clear to a prospective student that it starts in `2021`. The prototype designs mentioned in the ticket have formatted this data like `Starts September 2021`, which makes it much clearer to a prospective student when the course starts than `2021/22`. I'm not sure if we can say that all courses start in September - might be something to check with our BAs. But I don't think we should leave it as `Starts 2021/22`.

FergusMcGlynn (changes requested):

src/components/CourseDetails.js
FergusMcGlynn: BTW if you want to make that a separate subtask/PR of this ticket then no worries - I can accept this one pending those changes

mrgibbonsuoy: I'll fix it before it goes any further.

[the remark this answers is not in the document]
mrgibbonsuoy: Yes it does, we just put the calendar at this stage, but we'll have to merge the details as per Jordans design at some point.

src/tests/components/CourseDetails.test.js
mrgibbonsuoy: This was something that I was going to look at.  I think I'll deal with it now along with your above comment.

src/components/CourseDetails.js
RichardAllum: Also sounds good, though I think it's worth saying that `startDateTag` and other tags could sit inside `CourseDetails` while they're so simple.

FergusMcGlynn: Agreed - they can be `private` (i.e. non-exported) component(s) in `CourseDetails.js`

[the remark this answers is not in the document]
RichardAllum: Based on our current stories we have separate tags for length and start date, that's why calendar was chosen for start date.

FergusMcGlynn (approved):

src/components/CourseDetails.js
```diff
@@ -0,0 +1,32 @@
...
+    return (
+        <TagList>
+            <StartDateTag yearOfEntry={course.yearOfEntry} />
+        </TagList>
+    );
+};
+
+const StartDateTag = ({ yearOfEntry }) => {
+    if (yearOfEntry) {
+        const startYear = yearOfEntry.slice(0, 4);
```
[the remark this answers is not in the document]
RichardAllum: If the format changes from the standard academic year it should also be with a different property name (e.g. if we change to 'September 2021') as we've got a standard academic year format, I agree it's not the best choice though.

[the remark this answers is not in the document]
mrgibbonsuoy: We had a bit of  a discussion and arrived at a similar conclusion. As always happy to put this in and refactor taking that observation on?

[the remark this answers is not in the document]
FergusMcGlynn: Isn't this field manually added into some `<meta>` tags on course web pages? I was more thinking of the case where someone manually enters `20/21` or `next year` or something unexpected like that. I was wondering if we should be a tad defensive and if `yearOfEntry.slice(0, 4)` doesn't result in an integer then return null? e.g.
```
const StartDateTag = ({ yearOfEntry }) => {
    const startYear  = yearOfEntry?.slice(0, 4);
    if (Number.isInteger(startYear)) {
        return <Tag ... />
    }
    return null;
}
```
But it's not a show-stopper - I'm fine with proceeding as is if you're confident that we can rely on the data coming out of the Courses API.

PR #48, opened 2020-10-09
CFIN-326 add award tag

ncn1:


src/tests/components/CourseDetails.test.js
```diff
@@ -3,6 +3,19 @@ import React from "react";
...
+        const exampleCourse = {
+            award: "BA (Hons)",
+        };
+
+        render(<CourseDetails course={exampleCourse} />);
+
+        const tags = screen.getAllByRole("listitem");
+
+        expect(tags[0]).toHaveTextContent("BA (Hons)");
+        expect(within(tags[0]).getByTestId("tag-icon")).toHaveClass("c-icon--bank");
```
FergusMcGlynn, line 16: Probably don't need to use `getAllByRole` here given that there's only one `listitem` role?
```suggestion
        const tag = screen.getByRole("listitem");

        expect(tag).toHaveTextContent("BA (Hons)");
        expect(within(tag).getByTestId("tag-icon")).toHaveClass("c-icon--bank");
```

src/components/CourseDetails.js
```diff
@@ -7,11 +7,20 @@ import PropTypes from "prop-types";
...
         <TagList>
+            <AwardTag award={course.award} />
             <StartDateTag yearOfEntry={course.yearOfEntry} />
         </TagList>
     );
 };
 
+const AwardTag = ({ award }) => {
+    if (award) {
+        return <Tag icon="bank" mainText={award} />;
```
FergusMcGlynn, line 18: Minor thing - I expect the icon in the designs is `university` rather than `bank`, even though they look very similar (might even be the same?). Using `university` might be better in case future versions of Font Awesome change these icons.

src/components/CourseDetails.js
```diff
@@ -21,6 +30,10 @@ const StartDateTag = ({ yearOfEntry }) => {
     return null;
 };
 
+AwardTag.propTypes = {
+    award: PropTypes.string,
+};
```
FergusMcGlynn, line 35: Possibly worth moving this directly below the `AwardTag` function? I think in the rest of the application code we have the `propTypes` directly after the function that it relates to.

FergusMcGlynn (approved):

src/tests/components/CourseDetails.test.js
ncn1: Corrected in Mark's branch, so won't do here.

src/components/CourseDetails.js
ncn1: Also now done in Mark's branch.

src/tests/components/CourseDetails.test.js
FergusMcGlynn: Mark's branch has been merged (I think). Also isn't this a separate test from the one he wrote?

PR #49, opened 2020-10-09
CFIN 232 make length of course visible

mrgibbonsuoy:


FergusMcGlynn:
This PR seems to have all of the changes from CFIN-326 - is that intentional?

mrgibbonsuoy:
Nicola and I were working on it at the same time, as your review so I made the change as we were there. I appreciate it's not great in one sense but in view of the short cycle we thought let's do it now.

FergusMcGlynn:
So is this PR for both CFIN-232 AND CFIN-326?

src/components/CourseDetails.js
```diff
@@ -21,10 +31,26 @@ const StartDateTag = ({ yearOfEntry }) => {
...
     yearOfEntry: PropTypes.string,
 };
 
+LengthCourseTag.propTypes = {
+    length: PropTypes.string,
+};
+
 CourseDetails.propTypes = {
     course: COURSE_MODEL,
 };
```
FergusMcGlynn, line 56: To be consistent with the rest of the application code, I think each `propType` declaration should go directly beneath the function that it relates to

FergusMcGlynn (approved):

src/components/CourseDetails.js
mrgibbonsuoy: I'll change that now. We had stylistically grouped them that way. But I'll make that change now.

src/tests/components/Course.test.js
```diff
@@ -4,6 +4,7 @@ import { Course } from "../../components/Course";
 describe("Course", () => {
     it("displays the course title", () => {
         const course = {
+            length: "",
             title: "Mathematics",
             liveUrl: "",
         };
```
RichardAllum, line 10:
```suggestion
        const course = {
            title: "Mathematics",
            liveUrl: "",
        };
```
I don't think we need to have the length specified in this test - it might confuse people in future

RichardAllum (approved):

src/tests/components/Course.test.js
mrgibbonsuoy: When I'd summitted the code initially there was a failure in testing. This was one of the areas identified. I have just re run format and fc just to cover both off and all the tests and it works with the line removed so I'll take your suggestion as you are correct.

PR #57, opened 2020-10-21
CFIN-224 Wrap CourseDetails in description div for padding

RichardAllum:
This just adds padding to the content of our Course details, examples can be seen at https://york.ac.uk/pattern-library/css-components/listings.html

### DIFF
--- a/src/components/Course.js
+++ b/src/components/Course.js
@@ -15,7 +15,9 @@ const Course = ({ course }) => {
         <li className="c-listings-item ">
             <a className="c-listings-item__link" href={course.liveUrl}>
                 <h2 className="c-listings-item__title">{title(course)}</h2>
-                <CourseDetails course={course} />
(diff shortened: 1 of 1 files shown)


FergusMcGlynn:
I know this has been merged now - but I just wanted to flag up that there are still some noticeable differences between [dev](https://courses.dev.app.york.ac.uk/) and [live](https://courses.app.york.ac.uk/). 
- dev still looks a bit more squashed - the padding seems to be between the title and the list break line, but there's no (or little) padding between the title and the tags
- the tags appear smaller in dev than in live
- also the link doesn't seem to cover the whole search result anymore - the padded bits above and below the result aren't part of the link

None of these are show-stoppers for me - but I just wanted to bring them up in case they are unintentional changes, as we said that this ticket ought to not affect the UI other than making the whole search result clickable.

PR #58, opened 2020-10-22
CFIN-357: Make "dev" Course Search use "dev" Courses API

FergusMcGlynn:
This turned out to be fairly straightforward once I realised that the environment variables need to be set in AWS Lambda (by Serverless) and found an [appropriate Serverless plugin](https://www.serverless.com/plugins/serverless-dotenv-plugin) to do the loading of the environment files.

Next.js only really understands 2 environments - development and production. In development mode (which you get by running `next dev` locally) it does babel transpiling, hot reloading, webpack on the fly etc, things that we don't really want it to do in AWS Lambda. So when we deploy the Next.js application to AWS Lambda, it really needs to be in production mode. But as a result, Next will only read environment variables from `.env.production`, and can't be told to look elsewhere for environment variables. However, you can set environment variables on the Lambda itself, using Serverless, and because the Next app is running inside the Lambda then it will honour those environment variables. That's how this solution works.

I will outline this in the README as part of the documentation sub-task.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -38,7 +38,7 @@ jobs:
 
       - name: Deploy - dev
         if: github.ref == 'refs/heads/dev'
-        run: npm run deploy
+        run: npm run deploy:dev
         env:
           AWS_ACCESS_KEY_ID: ${{ secrets.DEV_AWS_ACCESS_KEY_ID }}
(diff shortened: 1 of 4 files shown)


ncn1 (approved):

ncn1:
> This turned out to be fairly straightforward ...
Good detective work!

ce666-york (approved):

PR #59, opened 2020-10-26
CFIN-364: Add documentation about switching between API versions

FergusMcGlynn:


ncn1 (approved):

ce666-york (approved):
Great documentation!

As an aside I'm starting to think reviewing the `README` and seeing what can be moved into wiki pages might be worthwhile soon, as the length could be a little overwhelming for someone new to the project.

PR #60, opened 2020-10-27
CFIN-374 Show a maximum of 20 course search results

mrgibbonsuoy:
A few tests have been updated to allow us to test different parts of the Courses API url being called. 

There is an option to put COURSES_API_MAX_RESULTS in a single .env file as it's a global value, however we have used the separate files for this PR, has anyone any preferences either way?

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -1,1 +1,2 @@
 COURSES_API_BASEURL=https://api.v1.courses.dev.app.york.ac.uk/courses
+COURSES_API_MAX_RESULTS=20
--- a/src/tests/pages/index.test.js
+++ b/src/tests/pages/index.test.js
@@ -82,7 +82,7 @@ describe("getServerSideProps", () => {
         expect(calledUrl).toContain("search=english");
     });
 
-    it("calls the Courses API with a max value", async () => {
+    it("calls the Courses API with a maximum number of results to return", async () => {
         await getServerSideProps(contextWithSearchTerm);
 
         expect(fetch).toHaveBeenCalledTimes(1);
--- a/src/tests/pages/index.test.js
+++ b/src/tests/pages/index.test.js
@@ -6,6 +6,7 @@ beforeEach(() => {
 });
 
 const emptyContext = { query: {} };
+const contextWithSearchTerm = { query: { search: "english" } };
 
 describe("App", () => {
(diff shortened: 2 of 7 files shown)


.env.development
```diff
@@ -1 +1,2 @@
 COURSES_API_BASEURL=https://api.v1.courses.dev.app.york.ac.uk/courses
+COURSES_API_MAX_RESULTS=20
```
FergusMcGlynn, line 2: I think of this as a app-wide global, rather than specific to any one environment, so I'd lean towards putting it in `.env` (all environments) rather than duplicating it in each specific environment

src/tests/pages/index.test.js
```diff
@@ -45,21 +46,49 @@ describe("getServerSideProps", () => {
...
+    it("calls the Courses API with a search term", async () => {
+        await getServerSideProps(contextWithSearchTerm);
+
+        expect(fetch).toHaveBeenCalledTimes(1);
+
+        const calledUrl = fetch.mock.calls[0][0];
+        expect(calledUrl).toContain("search=english");
+    });
+
+    it("calls the Courses API with a max value", async () => {
```
FergusMcGlynn, line 85: Very minor thing, but it's not immediately obvious to a noob what a `max value` is, so a tiny bit of clarification might help?
```suggestion
    it("calls the Courses API with a maximum number of results to return", async () => {
```

ncn1 (approved):

.env.development
ce666-york: I already mentioned to Mark that I don't have a strong preference on this, I think it's fine to work with and understand either way.

- If it's in a global `.env` file then it only needs to be updated in one place
- If it's in environment-specific `.env` files then config for an environment is all stored in one place

ce666-york (approved):

RichardAllum (approved):

.env.development
mrgibbonsuoy: Ferg and I have had a chat about this. We're going to leave it as it is, and then I'll introduce a ticket into the back log to look at the best way to store environment files going forward. 
There is a case for having them in a constants file for example and merging .env.production and .env.development in future, however at present we'll merge this branch.

PR #62, opened 2020-10-28
CFIN 361 number of search results

mrgibbonsuoy:
This PR is to add in the numberOfMatches results and to refactor unit tests to handle this new value. Other work will be required to display the results.


mrgibbonsuoy:
Stopped due to duplication.

PR #63, opened 2020-10-29
CFIN 361 number of search results

mrgibbonsuoy:
This is a change to limit the number of results on the screen and inform the user how many results are returned. 

In the case that we are showing all results we will display the text "Showing all 5 results for Archaeology".

In the case that there are more matches than we are showing we display the text "Showing the top 20 results for English"

You can see this functionality in my personal sandbox at: https://d28f89qbw6jana.cloudfront.net/?search=english

### DIFF
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -13,7 +13,7 @@ import { COURSE_MODEL } from "../constants/CourseModel";
 import { PageHead } from "../components/PageHead";
 import { Search } from "../components/Search";
 
-const App = ({ isSuccessfulSearch, searchResults, numberOfResults, searchTerm }) => {
+const App = ({ isSuccessfulSearch, searchResults, numberOfMatches, numberOfResults, searchTerm }) => {
     return (
         <>
             <PageHead search={searchTerm} />
@@ -23,7 +23,11 @@ const App = ({ isSuccessfulSearch, searchResults, numberOfResults, searchTerm })
             <WrappedMainGrid>
                 <GridRow>
                     <GridBoxFull>
-                        <Search searchTerm={searchTerm} numberOfResults={numberOfResults} />
+                        <Search
+                            searchTerm={searchTerm}
+                            numberOfMatches={numberOfMatches}
+                            numberOfResults={numberOfResults}
+                        />
                     </GridBoxFull>
                 </GridRow>
                 <GridRow>
@@ -41,6 +45,7 @@ const App = ({ isSuccessfulSearch, searchResults, numberOfResults, searchTerm })
 App.propTypes = {
     isSuccessfulSearch: PropTypes.bool,
     searchResults: PropTypes.arrayOf(COURSE_MODEL),
+    numberOfMatches: PropTypes.number,
     numberOfResults: PropTypes.number,
     searchTerm: PropTypes.string,
 };
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -55,13 +55,13 @@ const getServerSideProps = async (context) => {
     try {
         const response = await fetch(courseSearchUrl);
         isSuccessfulSearch = response.ok;
-        searchResponseData = isSuccessfulSearch ? await response.json() : { results: [] };
+        searchResponseData = isSuccessfulSearch ? await response.json() : { numberOfMatches: 0, results: [] };
     } catch {
         isSuccessfulSearch = false;
-        searchResponseData = { results: [] };
+        searchResponseData = { numberOfMatches: 0, results: [] };
     }
 
-    return { props: { isSuccessfulSearch, searchResults: searchResponseData.results, searchTerm } };
+    return { props: { isSuccessfulSearch, searchResults: searchResponseData, searchTerm } };
 };
 
 export { App as default, getServerSideProps };
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -61,7 +61,14 @@ const getServerSideProps = async (context) => {
         searchResponseData = { numberOfMatches: 0, results: [] };
     }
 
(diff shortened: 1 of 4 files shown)


src/tests/components/Search.test.js
```diff
@@ -22,4 +22,18 @@ describe("Search", () => {
 
         expect(getByRole("textbox", { name: "Search" }).value).toEqual("French");
     });
+
+    it("informs the user that only a limited number of results are being shown when the number of matches is greater than the number of results", () => {
+        render(<Search searchTerm="Maths" numberOfMatches={25} numberOfResultsShown={23} />);
+
+        expect(screen.getByTestId("search-results-description")).toHaveTextContent(
```
ce666-york, line 29: Thoughts welcome on whether this could be better; I wondered if there was a more accessible way to do it with `getByRole`.

I originally tried using `getByText`, but this doesn't match due to the `<strong>` tag. [This issue](https://github.com/testing-library/dom-testing-library/issues/410) suggested using `data-testid` which then allows us to use `toHaveTextContent`, which does match the text.

ncn1 (approved):

ce666-york:
For info, we put a Cloudfront distribution in front of the sandbox because we thought we had an issue with trailing slashes that caused searches to fail, but I think it's actually due to the `v1` not being accounted for in the form action to submit a search. You can see this behaviour when using [API gateway](https://ldpvx4pvd0.execute-api.eu-west-1.amazonaws.com/v1) directly.

It would be nice if the app "just worked" when deploying to a sandbox, that's probably something to look at in future.

src/tests/components/Search.test.js
FergusMcGlynn: This is a good question. The React Testing Library [generally steers us away from `getByTestId`](https://testing-library.com/docs/guide-which-query), but in some cases its better than all other options. I'll have a think about whether there could be a more accessible way of testing (and marking up) the search description.

FergusMcGlynn:
> For info, we put a Cloudfront distribution in front of the sandbox because we thought we had an issue with trailing slashes that caused searches to fail, but I think it's actually due to the `v1` not being accounted for in the form action to submit a search. You can see this behaviour when using [API gateway](https://ldpvx4pvd0.execute-api.eu-west-1.amazonaws.com/v1) directly.
> 
> It would be nice if the app "just worked" when deploying to a sandbox, that's probably something to look at in future.

Yes, the problem is that all the links/paths/controls generated by Next.js are relative to `/`, and that won't work for API Gateway because it deploys the app to `/v1`. There might be a way of configuring the Next.js app to generate paths starting with `/v1`, but then that'll stop the app working when a domain name comes into play and makes the home path `/` again. I wondered if we could set up a domain name and SSL certificate e.g. `courses.sandbox.dev.app.york.ac.uk` in ESG Dev and then make use of that when we want to deploy the app to our sandboxes - but then I don't think you can use an SSL certificate from another AWS account. I agree that it'd be good to find a way of get this working in our sandboxes.

src/tests/components/Search.test.js
FergusMcGlynn: I've spent a bit of time thinking about this, and I can't think of a better way of testing this element, so I'm happy to go with what's here.

src/pages/index.js
```diff
@@ -55,13 +61,21 @@ const getServerSideProps = async (context) => {
...
+        searchResponseData = { numberOfMatches: 0, results: [] };
     }
 
-    return { props: { isSuccessfulSearch, searchResults: searchResponseData.results, searchTerm } };
+    return {
+        props: {
+            isSuccessfulSearch,
+            searchResults: searchResponseData.results,
+            numberOfMatches: searchResponseData.numberOfMatches,
+            numberOfResultsShown: searchResponseData.results.length,
```
FergusMcGlynn, line 75: Do we need to return `numberOfResultsShown`? It can be computed pretty simply from `searchResults`, so why do we need it?

mrgibbonsuoy: It was done in the index so that we could then just pass it along, it would then wrap everything up into a simple return.

FergusMcGlynn: I don't really understand what you mean I'm afraid :-/

ce666-york: > Do we need to return numberOfResultsShown?

As you say it can be computed so we don't need it. However it does mean that the calculation reads slightly more easily, and also we're not passing unnecessary information (the entire search results with all their details) into the `Search` component.

Another way to achieve both of these whilst avoiding the extra prop in `index.js` is to do the calculation in-line e.g.

```
<Search
    searchTerm={searchTerm}
    numberOfResultsShown={searchResults.length}
/>
```

Do you think this is worth changing?

FergusMcGlynn: Yes, that's what I was thinking. One less prop being passed to `App` makes it a little cleaner to my eyes. I'm also not a huge fan of passing around parameters that can be calculated from other parameters, unless the calculation is a lot of work. It could lead to an inconsistency where, for example, the page reports that there a 5 results, but 6 are presented. Very unlikely to happen here of course, although one of the unit tests does exactly that. I wasn't sure whether that was deliberate - to show that `numberOfResultsShown` is independent of `searchResults` - but they're not independent, so it's a little confusing in the test. 

Personally I'd change it unless we gain something `numberOfResultsShown` being an explicit prop of its own.

FergusMcGlynn: Oh wait - the test I just mentioned demonstrates `numberOfMatches` being independent of `numberOfResultsShown`, which is fine - they are genuinely independent. It still feels to me like the `numberOfResultsShown` in not necessary. But I don't mind if you prefer it the way that it is.

ce666-york: Happy to change it, will do a separate PR as this has been merged

PR #67, opened 2020-10-30
CFIN-244 increase the max results shown to 35

ce666-york:
A tiny update to show 35 course search results instead of 20.


src/tests/pages/index.test.js
```diff
@@ -99,7 +99,7 @@ describe("getServerSideProps", () => {
         expect(fetch).toHaveBeenCalledTimes(1);
 
         const calledUrl = fetch.mock.calls[0][0];
-        expect(calledUrl).toContain("max=20");
+        expect(calledUrl).toContain("max=35");
```
ncn1, line 102: Could the 35 not be replaced with COURSES_API_MAX_RESULTS, so tests don't have to be changed every time the number of results is updated?

ce666-york: It could, the reason we initially didn't was because we saw this test:
```
expect(calledUrl).toContain("https://test.courses.api.com");
```
so we stuck with using the expected value.

Changing it would mean this test becomes a "the course search shows whatever number of results the config says" rather than "the course search shows 35 results", but the current title kind of says the former anyway, so I'm inclined to change both this one and the URL test.

FergusMcGlynn: `.env.test` doesn't need to change to `35` just because `.env.production` has. It contains test values - so you can leave `.env.test` as `20` and not have to change the test. 

There is some value in verifying that the code is picking up values from correct environment, so I'd prefer to keep these values as literal `20` or `35` rather than `COURSES_API_MAX_RESULTS`. I feel happier when tests verify concrete things personally.

ncn1: I think "the course search shows whatever number of results the config says" is actually a more valid test.

mrgibbonsuoy: Just off the topic slightly, but for visibility, I've raised CFIN-380 in the backlog to look at; some point in the future, creating a global store for this value. 
As Ferg has suggested placing the value in one location will simplify things in future.

mrgibbonsuoy (approved):

src/tests/pages/index.test.js
FergusMcGlynn: If we test for `process.env.COURSES_API_MAX_RESULTS` rather than `20`, and the environment isn't configured properly, isn't there a risk that `process.env.COURSES_API_MAX_RESULTS` will be `undefined`, which evaluates to the string `''`, and when testing for equality, the test will pass? I'd say in that case it'd be good if the test failed so that you know the application isn't picking up environment variables correctly.

ncn1: That sounds like it should be a separate test.  I'm OK with a literal here, but if we happened to be checking this number out multiple times, I'd advocate one test to check it's got the number for the right environment, and the rest just reference the variable, to avoid lots of changes to the tests, as a general principle.

ChristianBeddows (approved):

src/tests/pages/index.test.js
ce666-york: All valid points, I like Nicola's suggestion of splitting them out into two separate tests.

To be honest the most important test is probably a functional one that proves that Course Search shows 35 results :slightly_smiling_face: Anyone want to investigate Cypress? :wink:

mrgibbonsuoy: More than happy to keep on with it Chris.

ncn1 (approved):

FergusMcGlynn (approved):

PR #70, opened 2020-11-03
CFIN-130 Add Google Analytics

ChristianBeddows:
This solution is based on:

- https://wiki.york.ac.uk/display/ittechdocs/Adding+Google+Analytics+to+a+web+application
- https://medium.com/@austintoddj/using-google-analytics-with-next-js-423ea2d16a98

I played around with this and tried a few other things, which didn't work - happy to hear any suggestions


src/utils/analytics.js
```diff
@@ -0,0 +1,21 @@
+import ReactGA from "react-ga";
```
RichardAllum, line 1: Does including the GA snippet directly in `pageHead.js` not work?

ChristianBeddows: It didn't when I tried it, threw a few syntax errors and prettier made it look horrible

src/pages/index.js
```diff
@@ -12,8 +12,13 @@ import { CourseSearchResults } from "../components/CourseSearchResults";
 import { COURSE_MODEL } from "../constants/CourseModel";
 import { PageHead } from "../components/PageHead";
 import { Search } from "../components/Search";
+import { logPageView } from "../utils/analytics";
 
 const App = ({ isSuccessfulSearch, searchResults, numberOfMatches, searchTerm }) => {
+    useEffect(() => {
+        logPageView();
+    }, []);
```
ce666-york, line 20: I've no problem with using a plugin if that's the best way to manage GA, but this bit worries me - I don't see why we  should have to concern ourselves with the actual logging? Especially as this isn't a single-page application.

src/utils/analytics.js
ce666-york: I think this would be the ideal if we can get it to work - worth a bit more wrangling? Happy to pair up to take a look!

FergusMcGlynn:
Along similar lines to what Chris and Richard have already said - would it be simpler if we followed the `New advice (November 2019)` section from [the wiki page](https://wiki.york.ac.uk/display/ittechdocs/Adding+Google+Analytics+to+a+web+application) rather than going with the React approach? Being server side rendered, that seems like the tidier approach. I'll admit that I don't really know much about Google Analytics though so happy to be told otherwise.

ChristianBeddows:
There's a [known issue](https://github.com/vercel/next.js/issues/9070) where next.js' hydration process will duplicate async script tags in the head. It appears that there is no easy workaround for this, however next already have an [open PR for this](https://github.com/vercel/next.js/pull/17497).

public/js/analytics.js
```diff
@@ -0,0 +1,6 @@
+window.dataLayer = window.dataLayer || [];
+function gtag() {
+    dataLayer.push(arguments);
+}
+gtag("js", new Date());
+gtag("config", "UA-1621853-1");
```
FergusMcGlynn, line 6: In [the wiki page](https://wiki.york.ac.uk/display/ittechdocs/Adding+Google+Analytics+to+a+web+application), the first set of instructions references `UA-1621853-1`, but in the "New advice (November 2019)", the id specified is `UA-149035954-1`. Do we know that this ID is the one we should be using?

ce666-york (approved):

public/js/analytics.js
ChristianBeddows: `UA-1621853-1` is the ID for the main UoY account used by marketing, which is what we'll be using for the new course search. 
`UA-149035954-1` is the ID for the ESG account, which we have more control over, but can't be accessed by marketing.

FergusMcGlynn: Ah, great - thanks! I might add that information to the wiki page as it seems quite important.

FergusMcGlynn (approved):

RichardAllum (approved):

PR #71, opened 2020-11-03
Add workflow to check the package size

ce666-york:
This PR is a suggestion, so thoughts would be very welcome! :thought_balloon: 

Inspired by #65 where I upgraded the application to Next 10 which unwittingly increased the package size, and #69 from Ferg which excluded the relevant dependency to shrink it back down, I thought it might be a good idea to get more visibility of our package size. This PR adds a workflow that uses the [compressed-size-action](https://github.com/marketplace/actions/compressed-size-action) (from the maintainers of preactjs) to do exactly that.

I recreated the above PRs in a demo repo to see what it'd look like, you can check them out [here](https://github.com/university-of-york/ce666-demo-repo/pull/6) and [here](https://github.com/university-of-york/ce666-demo-repo/pull/7) (and [here](https://github.com/university-of-york/ce666-demo-repo/pull/8) for one that shows minimal changes).

One thing to consider is that email notifications are sent for comments (unless you turn them off in your notification settings).

**Note**: The check doesn't run successfully in this PR because the action works by comparing the current branch against the target branch, and the `package` command that I'm using here doesn't yet exist in the target branch.

### DIFF
--- a/.github/workflows/report-package-size.yml
+++ b/.github/workflows/report-package-size.yml
@@ -26,5 +26,5 @@ jobs:
           compression: none
           minimum-change-threshold: 1024
         env:
-          NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
           AWS_ACCOUNT_ID: ${{ secrets.DEV_AWS_ACCOUNT_ID }}
--- a/.github/workflows/report-package-size.yml
+++ b/.github/workflows/report-package-size.yml
@@ -1,4 +1,4 @@
-name: Application Size
+name: Report Package Size
 
 on: [pull_request]
 
@@ -17,7 +17,7 @@ jobs:
           scope: "@university-of-york"
           always-auth: true
 
-      - name: Check the application's file size
+      - name: Check and report changes in the application's package size
         uses: preactjs/compressed-size-action@e2bfc4b346d78f9a3c7e8a2e22bcbd285a25fafa
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/report-package-size.yml
+++ b/.github/workflows/report-package-size.yml
@@ -28,4 +28,3 @@ jobs:
         env:
           NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
           AWS_ACCOUNT_ID: ${{ secrets.DEV_AWS_ACCOUNT_ID }}
-
--- a/.github/workflows/application-size.yml
+++ b/.github/workflows/application-size.yml
@@ -1,0 +1,31 @@
+name: Application Size
+
+on: [pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the application's source code
+        uses: actions/checkout@v2
+
+      - name: Install node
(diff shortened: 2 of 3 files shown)


.github/workflows/report-package-size.yml
```diff
@@ -0,0 +1,30 @@
...
+      - name: Install node
+        uses: actions/setup-node@v1
+        with:
+          node-version: 12
+          registry-url: https://npm.pkg.github.com
+          scope: "@university-of-york"
+          always-auth: true
+
+      - name: Check and report changes in the application's package size
+        uses: preactjs/compressed-size-action@e2bfc4b346d78f9a3c7e8a2e22bcbd285a25fafa
```
FergusMcGlynn, line 21: Do we want to pin to a specific commit? The [documentation](https://github.com/preactjs/compressed-size-action#usage) suggests:
```suggestion
        uses: preactjs/compressed-size-action@v2
```

.github/workflows/report-package-size.yml
```diff
@@ -0,0 +1,30 @@
...
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          build-script: package
+          pattern: .serverless/*.zip
+          compression: none
+          minimum-change-threshold: 1024
```
FergusMcGlynn, line 27: Personally, I'd be interested in changes smaller than 1Mb - although I'm struggling to quantify exactly where my interest would start. Maybe 256kb? I want the deployment to be as small as it can reasonably be. 50Mb is the Lambda hard limit (I believe) and it'd be good to stay way below that (~16Mb at present). Having said that, I appreciate this is mainly for catching big jumps like what happened with Next.js v10, so I don't mind if we want to stick with 1Mb for now and see how we get on.

```suggestion
          minimum-change-threshold: 256
```

FergusMcGlynn (approved):
Great idea - I love this! 🥇 👍 😃

.github/workflows/report-package-size.yml
ce666-york: Commit pinning is recommended in our [guide on using third party actions](https://wiki.york.ac.uk/display/ittechdocs/Use+of+Third+Party+GitHub+Actions) - admittedly I was the one who wrote this guide so happy to start a discussion 🙂

.github/workflows/report-package-size.yml
ce666-york: Just double-checked the documentation and [this value is in bytes](https://github.com/marketplace/actions/compressed-size-action#increasing-the-required-threshold) - so any change above 1KiB is flagged!

Thinking about it, this value is basically a threshold for "do I need to think about this PR more" - so upping it to 256 (i.e. 262144) might be sensible anyway.

.github/workflows/report-package-size.yml
FergusMcGlynn: Ah, fair enough, I had forgotten about that. Thanks for the reminder!

.github/workflows/report-package-size.yml
FergusMcGlynn: I did not spot that the value was in bytes - thanks for checking. I don't know why I assumed it was in kb. In that case, I personally don't mind it being left how it is at 1kb - that might help us get a feeling for how our changes impact the bundle size more regularly than 256kb would.

ncn1 (approved):

PR #72, opened 2020-11-04
CFIN-384: Show no search results when no search terms entered

FergusMcGlynn:
This PR includes changes to ensure that no search results, and no search description text (e.g. "showing top n results"), are presented when there are no search terms. A future PR will add functionality to display a message to the user stating that they didn't enter any search terms.

[Try it out in my sandbox](https://courses.ferg.dev.app.york.ac.uk/)

### DIFF
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -32,7 +32,11 @@ const App = ({ isSuccessfulSearch, searchResults, numberOfMatches, searchTerm })
                 </GridRow>
                 <GridRow>
                     <GridBoxFull>
-                        <CourseSearchResults isSuccessfulSearch={isSuccessfulSearch} searchResults={searchResults} />
+                        <CourseSearchResults
+                            isSuccessfulSearch={isSuccessfulSearch}
+                            searchResults={searchResults}
+                            searchTerm={searchTerm}
+                        />
                     </GridBoxFull>
                 </GridRow>
             </WrappedMainGrid>
@@ -50,7 +54,11 @@ App.propTypes = {
 };
 
 const getServerSideProps = async (context) => {
-    const searchTerm = context.query.search || "maths";
+    const searchTerm = context.query.search;
+
+    if (!searchTerm) {
+        return { props: {} };
+    }
 
     const courseSearchUrl = `${process.env.COURSES_API_BASEURL}?search=${searchTerm}&max=${process.env.COURSES_API_MAX_RESULTS}`;
 
--- a/src/tests/components/CourseSearchResults.test.js
+++ b/src/tests/components/CourseSearchResults.test.js
@@ -8,14 +8,14 @@ describe("CourseSearchResults", () => {
             { title: "Physics", liveUrl: "http://foo.baz" },
         ];
 
-        render(<CourseSearchResults isSuccessfulSearch searchResults={searchResults} />);
+        render(<CourseSearchResults isSuccessfulSearch searchResults={searchResults} searchTerm="foobar" />);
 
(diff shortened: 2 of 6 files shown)


src/tests/components/CourseSearchResults.test.js
```diff
@@ -26,8 +26,19 @@ describe("CourseSearchResults", () => {
...
+        render(<CourseSearchResults isSuccessfulSearch searchResults={[]} searchTerm="foobar" />);
 
         expect(screen.getByText(/Sorry, your search did not return any results/)).toBeVisible();
     });
+
+    it.each`
+        description    | searchTerm
+        ${"blank"}     | ${""}
+        ${"null"}      | ${null}
+        ${"undefined"} | ${undefined}
```
ce666-york, line 38: I mean it's not as pretty as Spock, but it does the job! Nice find

src/tests/components/CourseSearchResults.test.js
```diff
@@ -26,8 +26,19 @@ describe("CourseSearchResults", () => {
...
+    `("displays nothing when user has entered $description search terms", ({ searchTerm }) => {
+        render(<CourseSearchResults searchTerm={searchTerm} />);
+
+        expect(screen.queryByText(/./)).not.toBeInTheDocument();
```
ce666-york, line 42: This regex casts quite a large net - might it be better to check for the lack of an element with the `search-results-description` `data-testid`?

src/pages/index.js
```diff
@@ -50,7 +54,11 @@ App.propTypes = {
 };
 
 const getServerSideProps = async (context) => {
-    const searchTerm = context.query.search || "maths";
+    const searchTerm = context.query.search;
+
+    if (!searchTerm) {
+        return { props: {} };
+    }
```
ce666-york, line 61: It's certainly a niggle seeing the same statement three times across this PR, but I'm not convinced any efforts to reduce duplication will make things simpler in this case...

ce666-york (approved):
Looks good, worth having a discussion about the regex but happy to approve 👍

src/tests/components/CourseSearchResults.test.js
FergusMcGlynn: I agree that the syntax isn't as nice as Spock :-/ Also, I can't claim to have done much detective work -[ it's relatively well documented in the Jest webpages](https://jestjs.io/docs/en/api.html#testeachtablename-fn-timeout).

src/pages/index.js
FergusMcGlynn: The upcoming PR is going to change it slightly - to discriminate between `searchTerm === ""` (the user has conducted a search with no search term) and `searchTerm === null / undefined` (no search has been conducted), so that might reduce the duplication somewhat.

src/tests/components/CourseSearchResults.test.js
FergusMcGlynn: The [React Testing Library guidelines](https://testing-library.com/docs/guide-which-query/) advise prioritising querying by text over querying by test id. Querying by text is more focused on testing what the user sees than querying by test id. 

I agree that this test is saying "I don't want to see any text whatsoever in this component for a blank search" - but to my mind that reflects the current app requirements.

ce666-york: Yup I agree with that, I think I wanted to flag up that this test will almost certainly break in future and need to be made more specific. I'm fine with fixing it at that point though.

ncn1 (approved):

PR #73, opened 2020-11-05
CFIN-383: Show "no results" message to user when they enter no search terms

FergusMcGlynn:
This solution relies on discriminating between `searchTerm === undefined` (meaning no search has been conducted yet) and `searchTerm === ""` (meaning an empty search has been conducted). I thought that was a bit of a subtle distinction so I've introduced a couple of helper functions to make these scenarios explicit.

[Try it out in my sandbox](https://courses.ferg.dev.app.york.ac.uk/)

### DIFF
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -12,6 +12,7 @@ import { CourseSearchResults } from "../components/CourseSearchResults";
 import { COURSE_MODEL } from "../constants/CourseModel";
 import { PageHead } from "../components/PageHead";
 import { Search } from "../components/Search";
+import { emptySearchConducted, noSearchConducted } from "../utils/searchTerms";
 
 const App = ({ isSuccessfulSearch, searchResults, numberOfMatches, searchTerm }) => {
     return (
@@ -56,11 +57,11 @@ App.propTypes = {
 const getServerSideProps = async (context) => {
     const searchTerm = context.query.search;
 
-    if (searchTerm === null || searchTerm === undefined) {
+    if (noSearchConducted(searchTerm)) {
         return { props: {} };
     }
 
-    if (searchTerm === "") {
+    if (emptySearchConducted(searchTerm)) {
         return { props: { searchTerm, isSuccessfulSearch: true, searchResults: [], numberOfMatches: 0 } };
     }
 
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -56,8 +56,12 @@ App.propTypes = {
 const getServerSideProps = async (context) => {
     const searchTerm = context.query.search;
 
-    if (!searchTerm) {
+    if (searchTerm === null || searchTerm === undefined) {
         return { props: {} };
+    }
+
+    if (searchTerm === "") {
+        return { props: { searchTerm, isSuccessfulSearch: true, searchResults: [], numberOfMatches: 0 } };
     }
 
     const courseSearchUrl = `${process.env.COURSES_API_BASEURL}?search=${searchTerm}&max=${process.env.COURSES_API_MAX_RESULTS}`;
--- a/src/components/CourseSearchResults.js
+++ b/src/components/CourseSearchResults.js
@@ -5,7 +5,7 @@ import React from "react";
 import { Alert } from "@university-of-york/esg-lib-pattern-library-react-components";
 
 const CourseSearchResults = ({ isSuccessfulSearch, searchResults, searchTerm }) => {
-    if (!searchTerm) {
+    if (searchTerm === null || searchTerm === undefined) {
         return null;
     }
 
--- a/src/components/CourseSearchResults.js
+++ b/src/components/CourseSearchResults.js
@@ -3,9 +3,10 @@ import PropTypes from "prop-types";
 import { COURSE_MODEL } from "../constants/CourseModel";
 import React from "react";
 import { Alert } from "@university-of-york/esg-lib-pattern-library-react-components";
+import { noSearchConducted } from "../utils/searchTerms";
 
 const CourseSearchResults = ({ isSuccessfulSearch, searchResults, searchTerm }) => {
-    if (searchTerm === null || searchTerm === undefined) {
+    if (noSearchConducted(searchTerm)) {
         return null;
     }
 
--- a/src/tests/components/CourseSearchResults.test.js
+++ b/src/tests/components/CourseSearchResults.test.js
@@ -15,7 +15,7 @@ describe("CourseSearchResults", () => {
     });
 
     it("displays an appropriate message when the course search failed", () => {
-        render(<CourseSearchResults isSuccessfulSearch={false} searchTerm="foobar" />);
+        render(<CourseSearchResults isSuccessfulSearch={false} searchResults={[]} searchTerm="foobar" />);
 
         expect(screen.getByText(/Course search is currently unavailable. Please try again later/)).toBeInTheDocument();
         expect(screen.getByRole("link", { name: "contact IT Support" })).toBeInTheDocument();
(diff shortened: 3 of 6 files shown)


ncn1 (approved):
I think the helper methods make this really clear.

This may not be the place to address this, but noticed entering a search for 'fff' gives a message 'Showing all 0 results for fff' as well as the 'Sorry, your search did not return any results.' message, which makes it a bit inconsistent. Needs discussing, I think.

mrgibbonsuoy (approved):

FergusMcGlynn:
> I think the helper methods make this really clear.
> 
> This may not be the place to address this, but noticed entering a search for 'fff' gives a message 'Showing all 0 results for fff' as well as the 'Sorry, your search did not return any results.' message, which makes it a bit inconsistent. Needs discussing, I think.

Good point! Claire brought this up in her testing of CFIN-244, and we agreed that it needed to be resolved. I've added a sub-task to the parent of this ticket for resolving it, and I'm currently working on that right now.

src/pages/index.js
```diff
@@ -56,10 +57,14 @@ App.propTypes = {
 const getServerSideProps = async (context) => {
     const searchTerm = context.query.search;
 
-    if (!searchTerm) {
+    if (noSearchConducted(searchTerm)) {
         return { props: {} };
     }
 
+    if (emptySearchConducted(searchTerm)) {
+        return { props: { searchTerm, isSuccessfulSearch: true, searchResults: [], numberOfMatches: 0 } };
```
ce666-york, line 65: I'm starting to get a little uncomfortable with the number of return statements and duplicate props in this method - worth spending a bit of time seeing if it can be refactored into something nicer?

FergusMcGlynn: I see the multiple return statements as [guard clauses](https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html). If the alternative you have in mind is building up `props` and having a single return statement at the end, I'm not sure that would wind up much cleaner, although I'm happy to have a go. 

Could you elaborate on what you mean by "duplicate props"? The prop values aren't duplicated as far as I can see, although `props` is mentioned in each return statement - is that you mean?

FergusMcGlynn: I've had a play with refactoring `getServerSideProps` - [is this the kind of thing you mean](https://github.com/university-of-york/uoy-app-course-search/commit/9d640ff9a3ab3289b0f9e5b6ea2edf2768bf6487)? (I confess that it contains as many `return` statements as before, but fewer references to `props`).

ce666-york: Sure! I recognise the first two `return`s as guard clauses, and I don't believe we should be aiming for a single return statement.

I'm not comfortable with the generally weak support `PropTypes` (and JavaScript) gives us - the `props` structure isn't strongly enforced, so the more separate places we're using it the more chances to get it wrong in future. The tests should hopefully prevent some of this.

Another part of my issue is the duplicate `{ numberOfMatches: 0, results: [] }` in the `try`-`catch` - I've been pushed over a threshold to mention it now 🙂

I don't have the answer, and refactoring isn't necessarily it, which is why I'm suggesting spending a bit of time to see - it'd be nice to know we've at least given it some thought. You may have already, in which case that's fine.

ce666-york: I wrote the above before seeing your last comment, so I'll catch up on that

ce666-york: I'm not sure either of those are any clearer - what do you think? If you look at the `search` function in isolation there's no definition of the expected return structure, whereas before at least `props:` gives you that hint.

FergusMcGlynn: Yes, I'm tend to agree that it's not massively clearer. I think there might be some value in the `fetchSearchResults` function - as that is doing a fairly specific unit of work, and would make `getServerSideProps` shorter and a bit easier to read. 

I agree that my refactor doesn't address the point you make about how some of the structures that we're passing around feel a bit arbitrary. `PropTypes` is enforcing them to an extent, although we haven't made any of the props `required`, which would add an extra level of enforcement. We could make use of [default prop values](https://github.com/university-of-york/uoy-app-course-search/compare/CFIN-383...CFIN-383_tidyup), which could reduce some of the duplication.

I will have a play with `PropTypes` to whether it can provide a bit more comfort here.

Another idea (although a much bigger piece of work) would be to move to Typescript, which would probably remove the feeling of arbitrary structures being passed around unchecked.

FergusMcGlynn: My limited play with `PropTypes` this afternoon hasn't given me any great ideas about how to make returning of `props` more robust. Things we can do to make `PropTypes` checking more rigourous:
- Mark `PropType` properties as `isRequired`
  - there'd be some work involved if we set `searchTerm` as `isRequired` * 
- use [prop-types-exact](https://github.com/airbnb/prop-types-exact) to ensure that only the specified `props` are given and no others
  - this library hasn't been updated in 2 years, but might be fine

I tried using default props - and felt like it won't really give us anything much. There's only one line where we return the "default" props, and all it did was make that line simpler while adding a bit more guff to the `App()` function's signature.

PropTypes enforcement is limited to logging an error to the javascript console whenever a transgression occurs. This could easily be missed, e.g. when you're running tests and the error is logged at the start of output but all the tests still pass, or when you're running the app but don't have the console open. This doesn't help me feel like there's much type enforcement going on.

In summary, I'm not sure PropTypes is going to give us that strong typed feeling. If we want it, I think we ought to talk about using Typescript.

\* we won't be able to return the value `undefined` (illegal in Next.js) or `null` (warning from React Pattern Library search component) so might need another `props` value to denote when no search has taken place (which feels a little messy to me, but arguably it'd make the distinction more transparent)

FergusMcGlynn: If we want to reduce the number of times we explicitly return the `{ props: ... }` structure, we could try [something like this](https://github.com/university-of-york/uoy-app-course-search/commit/fde6279ab62397f86c47367590c02e4c0c30a9e3) - although I'm not sure it's that much cleaner to be honest.

ce666-york: Thanks for looking into this, I appreciate hearing your thoughts. I agree with the assessment that TypeScript would be the next step rather than spending more time on it from the current angle.

The subject of TypeScript has popped up a couple of times recently, if anyone feels strongly about it and wants to pull some information together I'm happy for us to look into it in more detail.

For this issue I think let's leave it where we are, and if you want to apply some of the refactoring that you saw value in then by all means :slightly_smiling_face:

ChristianBeddows: This does seem like quite a tricky thing to refactor. If we are going to have a lot of return statements within the one method, I think I prefer being verbose with the structure so I can see the differences between paths, rather than hiding structure away in helper methods.

However, the server side props method is getting quite large/complex. Perhaps moving some of the code into a function that performs the search would be good - and that function internally calls methods to validate the users search and return those specific values for the edge cases?

I think I have yet to see the real benefits of using PropTypes - so far I've seen it help with IDE inference but I can't think of any other ways in which it's particularly helped. I'd be happy to try out TypeScript at some point - I think there may be a reason other teams are keen on it.

ChristianBeddows (approved):

src/pages/index.js
FergusMcGlynn: Thanks Chris and Christian. On Christian's point about `getServerSideProps` becoming quite large - I agree. After this PR, I will create another PR that moves the block of code conducting the actual http request into its own private method. I think that'll make `getServerSideProps` more focused - but we can discuss that on the PR when it comes.

ce666-york (approved):

ncn1 (approved):

PR #75, opened 2020-11-09
CFIN-407 remove pattern library javascript

ce666-york:
[Lighthouse](https://developers.google.com/speed/pagespeed/insights/?url=https%3A%2F%2Fcourses.dev.app.york.ac.uk%2F) reports three issues of unused JS:

- The pattern library's `app.min.js`
- A Google APIs script (loaded via `app.min.js`)
- A Next.js static chunk

I think that the Next.js static chunk should be treated as a consequence of us choosing to use the framework, so it's not something we should be looking to customise unless we really need the benefits.

Therefore this PR focuses on removing `app.min.js`, which in itself is a fairly simple change but has some implications.

## Current pattern library functionality

It's been mentioned in the past that two parts of core pattern library functionality are delivered through JS:

- The sticky footer
- The "Show More" nav bar functionality

I also tried to consider any other global functionality, of which I found two examples:

- The cookie banner
- The Coronavirus global alert

The sticky footer doesn't appear to work particularly well with our app, as we also saw in the PMC where we decided to re-implement it. I suspect we'll end up with a similar solution in Course Search.

The "Show More" functionality is something we don't use now, but will almost certainly want to in future. Thankfully this is already implemented as a [React pattern library component](https://master.d16equcwnlw8rg.amplifyapp.com/?path=/story/menunavigation--with-main-and-secondary-list), and seemed to work perfectly well in Next.js when I did a quick test locally.

In the cases of the cookie banner and global alert, I think we would need to implement them in the React pattern library - I'd suggest this is worth doing (under separate stories). As things currently stand we don't include the global alert anyway, as it's delivered by a [separate script](https://www.york.ac.uk/static/globalalert/globalalert.js).

Therefore I think "new" work as a result of this change would be reintroducing the cookie banner.

## Impact

Given all the above, my suggestion is to exclude the pattern library JS, as it does improve the app's performance score significantly:

**Average Lighthouse Scores**

| With pattern library JS  | Without pattern library JS | Without pattern library JS and with [text compression enabled](https://jira.york.ac.uk/browse/CFIN-409) |
| ------------- | ------------- | ------------- |
| [57](https://developers.google.com/speed/pagespeed/insights/?url=https%3A%2F%2Fcourses.dev.app.york.ac.uk%2F)  | 72 | [88](https://developers.google.com/speed/pagespeed/insights/?url=https%3A%2F%2Fce666.dev.app.york.ac.uk) |

The above is the result of my personal thoughts and investigation, but I'd very much welcome more opinions and any other consequences I may have missed.

### DIFF
--- a/src/components/PageHead.js
+++ b/src/components/PageHead.js
@@ -16,9 +16,6 @@ const PageHead = ({ search }) => {
             <link rel="icon" type="image/x-icon" href="https://www.york.ac.uk/static/stable/img/favicon.ico" />
             <link rel="stylesheet" href="https://www.york.ac.uk/static/stable/css/styles.min.css" media="screen" />
 
-            <script src="https://www.york.ac.uk/static/stable/js/modernizr.min.js" />
-            <script src="https://www.york.ac.uk/static/stable/js/app.min.js" />
-
             <script src="//use.typekit.net/dvj8rpp.js" />
             <script language="application/javascript">Typekit.load();</script>
         </Head>


ce666-york:
I also did a quick test (in IE8 on Windows 7!) with/without the pattern library JS to check if it was solving any significant compatibility issues and there was no apparent difference in functionality (though in both cases it didn't look pretty!)

itspaulkelly:
Nick and I have had some conversations recently about how we need to improve our front-end performance and a lot of that relates to what's in app.min.js. There's a lot of legacy stuff in there that gets loaded even when it's not needed and it's having a performance impact across the board. Not quite sure when we'll get to that, but it's something we'd very much like to do.

RichardAllum (approved):

ncn1 (approved):

mrgibbonsuoy (approved):

FergusMcGlynn (approved):

PR #76, opened 2020-11-10
CFIN-412: Refactor option 1 - extract Courses API search into separate function

FergusMcGlynn:
This is the first of 2 alternative options for a refactor.

This one extracts the block of code that handles calling the Courses API  out of `getServerSideProps` and into its own private function in the same file. It's a fairly minimal change.

Do you prefer one or the other? Or neither?

### DIFF
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -54,25 +54,6 @@ App.propTypes = {
     searchTerm: PropTypes.string,
 };
 
-const searchForCourses = async (searchTerm) => {
-    const courseSearchUrl = `${process.env.COURSES_API_BASEURL}?search=${searchTerm}&max=${process.env.COURSES_API_MAX_RESULTS}`;
-
-    let isSuccessfulSearch;
-    let searchResponseData;
-    const defaultResponse = { numberOfMatches: 0, results: [] };
-
-    try {
-        const response = await fetch(courseSearchUrl);
-        isSuccessfulSearch = response.ok;
-        searchResponseData = isSuccessfulSearch ? await response.json() : defaultResponse;
-    } catch {
-        isSuccessfulSearch = false;
-        searchResponseData = defaultResponse;
-    }
-
-    return { isSuccessfulSearch, searchResponseData };
-};
-
 const getServerSideProps = async (context) => {
     const searchTerm = context.query.search;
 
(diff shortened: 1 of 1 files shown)


mrgibbonsuoy (approved):
I like this version. That's just my opinion :)

src/pages/index.js
```diff
@@ -54,37 +54,44 @@ App.propTypes = {
...
-
-    if (noSearchConducted(searchTerm)) {
-        return { props: {} };
-    }
-
-    if (emptySearchConducted(searchTerm)) {
-        return { props: { searchTerm, isSuccessfulSearch: true, searchResults: [], numberOfMatches: 0 } };
-    }
-
+const searchForCourses = async (searchTerm) => {
```
ce666-york, line 57: Any reason you decided to place this above `getServerSideProps`? Just thinking of Clean Code's "read down the page" advice

src/pages/index.js
```diff
@@ -54,37 +54,44 @@ App.propTypes = {
...
     const courseSearchUrl = `${process.env.COURSES_API_BASEURL}?search=${searchTerm}&max=${process.env.COURSES_API_MAX_RESULTS}`;
 
     let isSuccessfulSearch;
     let searchResponseData;
+    const defaultResponse = { numberOfMatches: 0, results: [] };
```
ce666-york, line 62: :+1:

ce666-york (approved):
I think I prefer this solution over the other, as `searchForCourses` doesn't feel big enough to me to be it's own module right now.

Either solution introduces the coupling I mentioned before of `searchForCourses` needing to comply with `App`'s `PropTypes` format, but refactoring does improve readability.

src/pages/index.js
FergusMcGlynn: I placed `searchForCourses` above `getServerSideProps` because it's a variable holding an arrow function, and I thought that variables needed to be declared before they can be used in javascript. I would prefer to declare it after `getServerSideProps` in line with Clean Code. I've just tried moving `searchForCourses` to after `getServerSideProps` - and it seems to work fine. But from what I've read, it shouldn't work unless I've "hoisted" `searchForCourses`... I'm going to spend a minute trying to work out what's going on. I agree that it would be preferable for the for private function to follow the public one.

RichardAllum (approved):
While I slightly prefer the other option I think this one would also be fine.

My reason for preferring option 2 is that it means HTTP requests are made and tested in one module, our tests for other files/modules can be simplified by mocking/stubbing that function instead. Since tests for HTTP requests can get awkward with things like `requests[0].params[x]` and `fetch.calls[0][x]` and other magic numbers, I think it's worth abstracting/hiding that away even if the function/module itself is small.

After writing this I realised we could just create a `getRequest` module/method like we did in the Proposals app that would also solve this problem, so I really don't have a strong preference.

src/pages/index.js
FergusMcGlynn: I haven't been able to figure out why it works, but it seems to work fine moving `searchForCourses` below `getServerSideProps`. Maybe Next/React does hoisting for you? Who knows... I've moved it as suggested.

FergusMcGlynn:
> After writing this I realised we could just create a `getRequest` module/method like we did in the Proposals app that would also solve this problem, so I really don't have a strong preference.

I think a `getRequest` module/method would be unnecessary at this stage. We only know of one case where this application needs to make a http request, searching for courses, so I think it makes sense for that one case to be a module/method of its own, with its own tests, and then it can be treated like a black box from then on.

PR #77, opened 2020-11-10
CFIN-412: Refactor option 2 - move Courses API search to new module

FergusMcGlynn:
This is the second of 2 alternative options for a refactor.

This one extracts the block of code that handles calling the Courses API out of getServerSideProps and into its own module, with its own tests, most of which have been moved over from `index.test.js`. This one makes `index.js` more concise - although it adds more code in total, because of the extra tests.

Do you prefer one or the other? Or neither?

### DIFF
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -54,6 +54,25 @@ App.propTypes = {
     searchTerm: PropTypes.string,
 };
 
+const searchForCourses = async (searchTerm) => {
+    const courseSearchUrl = `${process.env.COURSES_API_BASEURL}?search=${searchTerm}&max=${process.env.COURSES_API_MAX_RESULTS}`;
+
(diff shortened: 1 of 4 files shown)


ncn1 (approved):
Think I prefer this version.

RichardAllum (approved):
I slightly prefer this option because it puts the fetch requests in a function that can then be mocked, but have no strong preference.

Since realizing we could also use something like `getRequest()` for the same thing I really have no preference at all, both of these are good solutions.

ChristianBeddows:
I prefer this option - it just feels a little more tidy to me. I like the idea of having a 'service' that acts as an interface for retrieving info from the Courses API - all this functionality kept separate from the page component, which should just care mainly about its own layout.

ChristianBeddows (approved):

PR #78, opened 2020-11-10
CFIN 402 Create new style tags

ncn1:
For the purpose of being able to demo the styling etc, each course has 2 hard-coded tags for the extra details such as year abroad etc that are not currently being passed in the Courses API.

This has also necessitated some creativity with the tests, which I will comment on in situ

### DIFF
--- a/src/components/CourseDetails.js
+++ b/src/components/CourseDetails.js
@@ -1,7 +1,7 @@
 import React from "react";
 import { COURSE_MODEL } from "../constants/CourseModel";
 import { TagList } from "./TagList";
-import { ExtrasTag, Tag } from "./Tag";
+import { ArrayTag, Tag } from "./Tag";
 import PropTypes from "prop-types";
 
 const CourseDetails = ({ course }) => {
@@ -9,7 +9,7 @@ const CourseDetails = ({ course }) => {
         <TagList>
             <LevelAndAwardTag level={course.level} award={course.award} />
             <StartDateAndLengthTag yearOfEntry={course.yearOfEntry} length={course.length} />
-            <ExtrasTag
+            <ArrayTag
                 title="Extras"
                 content={[
                     { icon: "suitcase", text: "year abroad" },
@@ -17,7 +17,7 @@ const CourseDetails = ({ course }) => {
                     { icon: "mortar-board", text: "Distance Learning" },
                 ]}
             />
-            <ExtrasTag
+            <ArrayTag
                 title=""
                 content={[
                     { icon: "suitcase", text: "year abroad" },
--- a/src/components/CourseDetails.js
+++ b/src/components/CourseDetails.js
@@ -1,7 +1,7 @@
 import React from "react";
 import { COURSE_MODEL } from "../constants/CourseModel";
 import { TagList } from "./TagList";
-import { Tag } from "./Tag";
+import {ExtrasTag, Tag} from "./Tag";
 import PropTypes from "prop-types";
 
 const CourseDetails = ({ course }) => {
@@ -9,7 +9,17 @@ const CourseDetails = ({ course }) => {
         <TagList>
             <AwardTag level={course.level} award={course.award} />
             <StartDateAndLengthTag yearOfEntry={course.yearOfEntry} length={course.length} />
-            <Tag mainText="Extras" subText="something something" />
+            <ExtrasTag title="Extras" content={[
+              {icon:"suitcase", text:"year abroad"},
+              {icon:"plane", text:"year in industry"},
+              {icon:"mortar-board", text:"Distance Learning"},
+            ]} />
+            <ExtrasTag title="" content={[
+              {icon:"suitcase", text:"year abroad"},
+              {icon:"plane", text:"year in industry"},
+              {icon:"plane", text:"Distance Learning"},
+              {icon:"mortar-board", text:"Integrated Masters"}
+              ]} />
         </TagList>
     );
 };
@@ -20,7 +30,7 @@ CourseDetails.propTypes = {
 
 const AwardTag = ({ level, award }) => {
     if (level) {
-        return <Tag topIcon="mortar-board" mainText={level} subText={award} />;
+        return <Tag topIcon="mortar-board" title={level} subText={award} />;
     }
 
     return null;
@@ -33,7 +43,7 @@ AwardTag.propTypes = {
 const StartDateAndLengthTag = ({ yearOfEntry, length }) => {
     if (yearOfEntry) {
         const startYear = yearOfEntry.slice(0, 4);
-        return <Tag topIcon="clock-o" mainText={`Start ${startYear}`} subText={length} />;
+        return <Tag topIcon="clock-o" title={`Start ${startYear}`} subText={length} />;
     }
 
     return null;
--- a/src/components/CourseDetails.js
+++ b/src/components/CourseDetails.js
@@ -7,9 +7,8 @@ import PropTypes from "prop-types";
 const CourseDetails = ({ course }) => {
     return (
         <TagList>
-            <AwardTag award={course.award} />
-            <StartDateTag yearOfEntry={course.yearOfEntry} />
-            <LengthCourseTag length={course.length} />
+            <AwardTag level={course.level} award={course.award} />
+            <StartDateAndLengthTag yearOfEntry={course.yearOfEntry} length={course.length} />
         </TagList>
     );
 };
@@ -18,9 +17,9 @@ CourseDetails.propTypes = {
     course: COURSE_MODEL,
 };
 
-const AwardTag = ({ award }) => {
-    if (award) {
-        return <Tag icon="university" mainText={award} />;
+const AwardTag = ({ level, award }) => {
+    if (level) {
+        return <Tag icon="mortar-board" mainText={level} subText={award} />;
     }
 
     return null;
@@ -30,29 +29,17 @@ AwardTag.propTypes = {
     award: PropTypes.string,
 };
 
-const StartDateTag = ({ yearOfEntry }) => {
+const StartDateAndLengthTag = ({ yearOfEntry, length }) => {
     if (yearOfEntry) {
         const startYear = yearOfEntry.slice(0, 4);
-        return <Tag icon="calendar" mainText={`Starts ${startYear}`} />;
+        return <Tag icon="clock-o" mainText={`Start ${startYear}`} subText={length} />;
     }
 
     return null;
 };
 
-StartDateTag.propTypes = {
+StartDateAndLengthTag.propTypes = {
     yearOfEntry: PropTypes.string,
-};
-
-const LengthCourseTag = ({ length }) => {
-    if (length) {
-        return <Tag icon="clock-o" mainText={length} />;
-    }
-
-    return null;
-};
-
-LengthCourseTag.propTypes = {
-    length: PropTypes.string,
 };
 
 export { CourseDetails };
--- a/src/components/CourseDetails.js
+++ b/src/components/CourseDetails.js
@@ -22,7 +22,7 @@ const CourseDetails = ({ course }) => {
                 content={[
                     { icon: "suitcase", text: "year abroad" },
                     { icon: "plane", text: "year in industry" },
-                    { icon: "plane", text: "Distance Learning" },
+                    { icon: "mortar-board", text: "Distance Learning" },
                     { icon: "mortar-board", text: "Integrated Masters" },
                 ]}
             />
@@ -35,8 +35,11 @@ CourseDetails.propTypes = {
 };
 
 const LevelAndAwardTag = ({ level, award }) => {
-    const tag = level || award ? <Tag topIcon="mortar-board" title={level} subText={award} /> : null;
-    return tag;
+    if (level || award) {
+        return <Tag topIcon="mortar-board" title={level} subText={award} />;
+    }
+
+    return null;
 };
 
 LevelAndAwardTag.propTypes = {
--- a/src/components/CourseDetails.js
+++ b/src/components/CourseDetails.js
@@ -9,6 +9,7 @@ const CourseDetails = ({ course }) => {
         <TagList>
             <AwardTag level={course.level} award={course.award} />
             <StartDateAndLengthTag yearOfEntry={course.yearOfEntry} length={course.length} />
+            <Tag mainText="Extras" subText="something something" />
         </TagList>
     );
 };
--- a/src/components/CourseDetails.js
+++ b/src/components/CourseDetails.js
@@ -7,7 +7,7 @@ import PropTypes from "prop-types";
 const CourseDetails = ({ course }) => {
     return (
         <TagList>
-            <AwardTag level={course.level} award={course.award} />
+            <LevelAndAwardTag level={course.level} award={course.award} />
             <StartDateAndLengthTag yearOfEntry={course.yearOfEntry} length={course.length} />
             <ExtrasTag
                 title="Extras"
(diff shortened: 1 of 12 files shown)


src/components/CourseDetails.js
```diff
@@ -1,15 +1,31 @@
...
 const CourseDetails = ({ course }) => {
     return (
         <TagList>
-            <AwardTag award={course.award} />
-            <StartDateTag yearOfEntry={course.yearOfEntry} />
-            <LengthCourseTag length={course.length} />
+            <LevelAndAwardTag level={course.level} award={course.award} />
+            <StartDateAndLengthTag yearOfEntry={course.yearOfEntry} length={course.length} />
+            <ExtrasTag
+                title="Extras"
```
ncn1, line 13: Hard-coded tags

src/tests/components/CourseDetails.test.js
```diff
@@ -12,43 +13,123 @@ describe("Course details", () => {
...
+        const tag = screen.queryAllByRole("listitem");
+
+        if (tag.length === 0) {
+            expect((tag.length = 0));
+        } else {
+            expect(tag[0]).not.toHaveTextContent("undergraduate");
+            expect(tag[0]).not.toHaveTextContent("BA (Hons)");
+            expect(within(tag[0]).queryByTestId("tag-icon")).toBe(null);
+        }
+    });
```
ncn1, line 75: This is the fiddle to ensure the test works with the hard-coded tags, in which case there will be some tags returned, and also when there are no hard-coded tags, so queryAllByRole("listitem") returns an empty array. Will need refactoring once design is finalised.

mrgibbonsuoy (approved):

src/components/CourseDetails.js
```diff
@@ -1,15 +1,31 @@
...
+                    { icon: "plane", text: "year in industry" },
+                    { icon: "mortar-board", text: "Distance Learning" },
+                ]}
+            />
+            <ExtrasTag
+                title=""
+                content={[
+                    { icon: "suitcase", text: "year abroad" },
+                    { icon: "plane", text: "year in industry" },
+                    { icon: "plane", text: "Distance Learning" },
```
FergusMcGlynn, line 25: Should the "Distance Learning" icon in this tag be consistent with the "Distance Learning" icon in the previous `ExtrasTag`?

src/components/CourseDetails.js
FergusMcGlynn: Are these hard coded in order to demonstrate how they look? Presumably we don't want them present on every course result?

src/components/CourseDetails.js
```diff
@@ -18,40 +34,27 @@ CourseDetails.propTypes = {
...
 };
 
-const AwardTag = ({ award }) => {
-    if (award) {
-        return <Tag icon="university" mainText={award} />;
-    }
-
-    return null;
+const LevelAndAwardTag = ({ level, award }) => {
+    const tag = level || award ? <Tag topIcon="mortar-board" title={level} subText={award} /> : null;
```
FergusMcGlynn, line 38: I find this line a little hard to parse - could we expand it to make it easier to read?

src/components/Tag.js
```diff
@@ -1,18 +1,53 @@
...
 
-const Tag = ({ icon, mainText }) => (
-    <li className="c-tag-list__item">
-        <span className="c-tag">
-            <i aria-hidden className={`c-tag__icon c-icon c-icon--${icon}`} data-testid="tag-icon" />
-            {mainText}
-        </span>{" "}
-    </li>
+const Tag = ({ topIcon, title, subText }) => (
+    <div className="new-tag" role="listitem">
```
FergusMcGlynn, line 5: Is the class name `new-tag` for illustrative purposes? Fine if so, but if not I think it could be confusing.

src/pages/_app.js
```diff
@@ -0,0 +1,14 @@
...
+const MyApp = ({ Component, pageProps }) => {
+    return <Component {...pageProps} />;
+};
+
+MyApp.propTypes = {
+    Component: PropTypes.any,
+    pageProps: PropTypes.any,
+};
+
+export default MyApp;
```
FergusMcGlynn, line 14: This is new to me - how does this relate to the new tags design?

FergusMcGlynn: Also is `MyApp` the most appropriate name for this function? (It may well be - I don't understand the context enough to know - but it doesn't seem like a very descriptive name)

src/components/CourseDetails.js
ncn1: Yes - see introductory comment

src/components/CourseDetails.js
ncn1: Yes - see introductory comment

RichardAllum: They're just there for demoing, yeah. This PR isn't going into dev, so they'll be removed before that

src/components/CourseDetails.js
RichardAllum: Maybe! It's entirely seeing if things exist at the moment

ncn1: Would be less confusing, even though they are just here for demo purposes. I will change.

src/components/CourseDetails.js
ncn1: It's basically the equivalent of the Groovy elvis operator - if there is either a level or award, return something, else return nothing. It's a choice of this succinct one -liner or 6 lines of an if / else statement.

src/components/CourseDetails.js
```diff
@@ -18,40 +34,27 @@ CourseDetails.propTypes = {
...
+    return tag;
 };
```
RichardAllum, line 40:
```suggestion
const LevelAndAwardTag = ({ level, award }) => {
    if (level || award) {
        return <Tag topIcon="mortar-board" title={level} subText={award} />
    }
    return null;
};
```
per Ferg's comment above.

src/tests/components/CourseDetails.test.js
```diff
@@ -12,43 +13,123 @@ describe("Course details", () => {
...
+    it("does not display level and award tag when neither level or award is present", () => {
+        const exampleCourse = {};
+
+        render(<CourseDetails course={exampleCourse} />);
+
+        const tag = screen.queryAllByRole("listitem");
+
+        if (tag.length === 0) {
+            expect((tag.length = 0));
+        } else {
```
FergusMcGlynn, line 70: I don't understand this bit. Presumably `expect((tag.length = 0));` is doing nothing because it's assigning `0` to `tag.length`, and that's not something I'd expect to verify with `expect`. Can this be simplified to:
```suggestion
        if (tag.length > 0) {
```
(I don't know whether it can BTW)

Aside from the above - I think conditional checks in tests are a bit strange and I think we should avoid them. As you say, this is going to be refactored at some point. Although I wouldn't want it to be merged in to `dev` like this if we can avoid it.

RichardAllum (changes requested):
I think we'll want to take the `Tag/ArrayTag` components and commit them to CFIN-371 in a brand new commit once the styles are in the pattern library

src/components/CourseDetails.js
ncn1: I think as we use the original style a bit more, it will become familiar and no longer difficult to parse. When learning Groovy, I found most of it's shortcuts difficult to read, but came to really like them as the language became familiar.

src/pages/_app.js
ncn1: It's code that needs to be included so you can include a local css file. I blindly followed some instructions, so I don't actually understand it myself. Hopefully we can remove it once the new styles are in the pattern library.

src/tests/components/CourseSearchResults.test.js
```diff
@@ -10,8 +10,10 @@ describe("CourseSearchResults", () => {
 
         render(<CourseSearchResults isSuccessfulSearch searchResults={searchResults} />);
 
-        expect(screen.getByRole("link", { name: "Maths" })).toBeInTheDocument();
-        expect(screen.getByRole("link", { name: "Physics" })).toBeInTheDocument();
+        const headings = screen.getAllByRole("heading");
+
+        expect(headings[0]).toHaveTextContent("Maths");
+        expect(headings[1]).toHaveTextContent("Physics");
```
FergusMcGlynn, line 16: Any particular reason for selecting these as headings as opposed to links? I appreciate that they are both, but from a user's perspective I would have thought it's more important that they are links than that they are headings.

src/tests/components/TagList.test.js
```diff
@@ -1,23 +1,71 @@
...
+        expect(tags[0]).toHaveTextContent("year in industry");
+        expect(within(tags[0]).getByTestId("suitcase")).toHaveClass("c-icon--suitcase");
+        expect(within(tags[0]).getByTestId("plane")).toHaveClass("c-icon--plane");
+        expect(within(tags[0]).getByTestId("mortar-board")).toHaveClass("c-icon--mortar-board");
+
+        expect(tags[4]).not.toHaveTextContent("Extras");
+        expect(tags[4]).toHaveTextContent("year abroad");
+        expect(tags[4]).toHaveTextContent("year in industry");
+        expect(tags[4]).toHaveTextContent("Distance Learning");
+        expect(tags[4]).toHaveTextContent("Integrated Masters");
```
FergusMcGlynn, line 65: It's a bit weird to the reader how we're jumping to the 5th element in the `tags` array. I'm guessing that's because we have lists within lists, and `getAllByRole('listitem')` is fetching every list item. It makes me wonder whether this would be confusing for a screen reader.

A possible alternative might be to use `<dl>`, `<dt>` and `<dd>` for the sub lists. `<dt>` has the implicit role `term` and `<dd>` has the implict role `definition`, so we'd be able to `getAllByRole('term')` or `getAllByRole('definition')` and that might be less confusing to a screen reader.

src/tests/components/TagList.test.js
```diff
@@ -1,23 +1,71 @@
...
+        expect(within(tags[4]).getByTestId("suitcase")).toHaveClass("c-icon--suitcase");
+        expect(within(tags[4]).getByTestId("mortar-board")).toHaveClass("c-icon--mortar-board");
+        const planes = within(tags[4]).getAllByTestId("plane");
+        expect((planes.length = 2));
```
FergusMcGlynn, line 69: I don't think `plane` is a good name for a test id

src/components/CourseDetails.js
FergusMcGlynn: It's not clear what the operator precedence is in the original line - does the `||` take precedence, or does the `?`?

FergusMcGlynn: I prefer Richard's suggestion. It's not quite equivalent to the Groovy elvis operator because of the `||` - which is what justifies breaking it down.

src/tests/components/CourseDetails.test.js
```diff
@@ -12,43 +13,123 @@ describe("Course details", () => {
...
+        const exampleCourse = {
+            level: "undergraduate",
+        };
+
+        render(<CourseDetails course={exampleCourse} />);
+
+        const tag = screen.getAllByRole("listitem");
+
+        expect(tag[0]).toHaveTextContent("undergraduate");
+        expect(tag[0]).not.toHaveTextContent("BA (Hons)");
```
FergusMcGlynn, line 43: I find this line (`.not.toHaveTextContent("BA (Hons)")`) a bit arbitrary. It might as well say `.not.toHaveTextContent("banana")`. If the goal is to verify that only the text `undergraduate` is presented and no other text whatsoever then I suspect there's a more robust way to verify it.

FergusMcGlynn: Looks like you can use [a regular expression inside `toHaveTextContent`](https://github.com/testing-library/jest-dom#tohavetextcontent) - that could be a way of verifying that exactly `undergraduate` is matched and there's no other text:

```
expect(tag[0]).toHaveTextContent(/^undergraduate$/);
```

FergusMcGlynn (changes requested):
I'm not quite sure how much of this PR is demo-ing ideas and how much of it is intended to be merged into `dev`, so I feel conflicted about whether or not to `Request changes`. I'll do so, to reflect that I think some tests could be improved, in line with the comments I've left. Apologies if that's not keeping in the spirit of this PR!

FergusMcGlynn:
As an aside - it might be worth marking this PR as a "draft", if its purpose is to showcase some changes without requesting the full scrutiny of a code review. I'm still not entirely clear what level of reviewing is being requested.

src/components/CourseDetails.js
ncn1: Fergus, can I not persuade you to embrace the terniary operator for single if - then - else 's like this one? https://www.javascripttutorial.net/javascript-ternary-operator/ . I think nesting them gets very hard to follow, but I think single ones do remove a lot of verbosity.

ncn1: Sorry, did not see your reply to the other thread before adding my previous comment.

FergusMcGlynn: I'm fine with ternary operators for simple `if a then b else c` expressions, but this one is a bit more complicated than that. Is it saying 
```
level || (award ? <Tag ... /> : null)
```
or
```
(level || award) ? <Tag .../> : null
```
? It's not immediately obvious to me. Also I think one of the sub-expressions being JSX makes it a bit trickier to parse than a simple all-javascript ternary. 

Richard suggested an alternative which makes it really easy to parse, and it's only a couple of lines. I'm not going to die in a ditch over it, and if everyone else finds the ternary perfectly easy to understand then we can stick with it.

src/components/CourseDetails.js
ncn1: Annoyingly, if I put brackets around 'level || award' in the terniary version, Prettier removes them!

src/tests/components/CourseDetails.test.js
ncn1: Thanks - much better way of doing it.

src/tests/components/TagList.test.js
ncn1: Description lists are only supposed to have 1 <dd> element, whereas our Extras tab needs there to be multiple lines under the <dt>. So standard lists it needed to be. I agree it isn't great - I keep coming up against the need for a getByClass option, but this has so far been ruled out by the authors of the testing library.

src/tests/components/TagList.test.js
ncn1: It means it can be the same as the icon name. Again, class name would have been much simpler to use here, and negate the need for a test id.

src/tests/components/CourseSearchResults.test.js
ncn1: Yes, the name of the links now also includes all the text in the tags.

FergusMcGlynn: Isn't that potentially an accessibility problem? If a screen reader is going to read out everything for a link? This might need some thought.

ncn1: I think that should be left for CFIN-401 'Ensure accessibility of new tags' to check and address if necessary. I has assumed it was just a React testing library artifact.

FergusMcGlynn: Fair enough - so long as it's not forgotten

ncn1:
> I'm not quite sure how much of this PR is demo-ing ideas and how much of it is intended to be merged into `dev`, so I feel conflicted about whether or not to `Request changes`. I'll do so, to reflect that I think some tests could be improved, in line with the comments I've left. Apologies if that's not keeping in the spirit of this PR!

It was a bit of both, and the code review has been really helpful. But if you are now ok with the changes that have already been made and knowing that it will not go into dev like this (and that would require another pull request, anyway, as this is just merging into the story branch) would you be able to approve it now?

FergusMcGlynn (approved):

ncn1:
As Richard has pointed out, probably best not to merge these changes into the story branch, as all the hard-coding and tests that account for the hard-coding will need removing before merging into dev, so closing the PR without merging.

PR #81, opened 2020-11-11
CFIN-417 Improve Course Search response time

ce666-york:
This is the result of an investigation to see if we can improve the response time of the Course Search application (purely focused on the front-end - other investigations into the Courses API and Funnelback are happening separately). I looked at package size and memory/CPU allocation. 

In short, **there's no noticeable effect on the average performance score when altering package size or memory/CPU allocation**, but I think the changes are worth doing for other reasons (deployment time and cost) so I'm raising this PR.

## Package size

- I identified a number of files that are currently packaged up that aren't necessary in the lambda, and added exclude rules for them.
- I identified a few more of the larger `node_modules` runtime dependencies (>=~1MB) that we don't appear to be using, and added exclude rules for these too.

The average performance score after this (plus all of the previous optimisations) was **88.55**, which is essentially unchanged from the **88** reported previously.

## Memory/CPU allocation

I also investigated Memory/CPU allocation. You can see the [full stats](https://docs.google.com/spreadsheets/d/19WKDGuomFVoVoBKuYMaQiZPZmDF3Quk7M6Gqo-rYPN8/edit#gid=0), but in summary:

| | 128MB | 256MB | 1024MB | 3008MB |
| --- | --- | --- | --- | --- |
| **Average performance score** | 85.2 | 88.95 | 88.55 | 88.7 |

The results indicate that we can lower our costs by decreasing the lambda size to 256MB with no noticeable effect on performance.

### DIFF
--- a/serverless.yml
+++ b/serverless.yml
@@ -38,9 +38,20 @@ custom:
   domainName: ${env:DOMAIN_NAME, ''}
   sslCertificateArn: ${env:SSL_CERTIFICATE_ARN, ''}
 
-# Exclude next.js dependencies that are not needed at runtime
+# Exclude files and next.js dependencies that are not needed at runtime
 package:
   exclude:
+    - .*
+    - .github/**
+    - .idea/**
+    - docs/**
+    - src/**
+    - package.json
+    - package-lock.json
+    - README.md
+    - LICENSE
+    - setupTests.js
+    - jest.config.js
     - node_modules/@babel/**
     - node_modules/ally.js/**
     - node_modules/caniuse-lite/**
--- a/serverless.yml
+++ b/serverless.yml
@@ -85,6 +85,13 @@ package:
     - node_modules/tr46/**
     - node_modules/micromatch/**
     - node_modules/sharp/**
+    - node_modules/@ampproject/**
+    - node_modules/postcss-modules-parser/**
+    - node_modules/css-modules-loader-core/**
+    - node_modules/babel-plugin-react-css-modules/**
+    - node_modules/es-abstract/**
+    - node_modules/string.prototype.trimstart/**
+    - node_modules/string.prototype.trimend/**
 
 resources:
   Conditions:
--- a/serverless.yml
+++ b/serverless.yml
@@ -7,7 +7,7 @@ provider:
   region: eu-west-1
   # This role is required to deploy an application to AWS with an ESG Account. See https://wiki.york.ac.uk/display/AWS/AWS%3A+Github+Actions for details.
   cfnRole: arn:aws:iam::${env:AWS_ACCOUNT_ID}:role/GithubActionsDeploymentRole
-  memorySize: 1024
+  memorySize: 256
   environment:
     NODE_ENV: production
   stackTags:
--- a/serverless.yml
+++ b/serverless.yml
@@ -7,7 +7,7 @@ provider:
   region: eu-west-1
   # This role is required to deploy an application to AWS with an ESG Account. See https://wiki.york.ac.uk/display/AWS/AWS%3A+Github+Actions for details.
   cfnRole: arn:aws:iam::${env:AWS_ACCOUNT_ID}:role/GithubActionsDeploymentRole
-  memorySize: 256
+  memorySize: 1024
   environment:
     NODE_ENV: production
   stackTags:


FergusMcGlynn:
I'm not hugely surprised that changes in package size and CPU/memory don't make much difference when comparing "warm" invocations of Course Search - but I'd expect them to have some (possibly significant) difference for cold starts. Do we know whether Lighthouse loads a page a few times when computing its scores? It generates a few snapshots, which gives me the impression that it does multiple page loads, and if so, it might not be giving us a true sense of how a cold start affects user experience. I haven't tested it out for a while, but I have a vague feeling that when I played around with CPU/memory, the cold start delay was noticeable at 256Mb. But that was a while ago, and the app has changed a lot since then.

ce666-york:
> I'd expect them to have some (possibly significant) difference for cold starts

@FergusMcGlynn Good point! I think it's something we should investigate before merging this change.

ChristianBeddows:
> I think it's something we should investigate before merging this change.

Here's a graph from my Courses API results that I've put together to hopefully show the effect on memory allocation on the cold start time. I don't know how relevant this will be but perhaps it'll come in handy

Screenshot from 2020-11-12 11-09-06

ChristianBeddows (approved):

serverless.yml
```diff
@@ -7,7 +7,7 @@ provider:
   region: eu-west-1
   # This role is required to deploy an application to AWS with an ESG Account. See https://wiki.york.ac.uk/display/AWS/AWS%3A+Github+Actions for details.
   cfnRole: arn:aws:iam::${env:AWS_ACCOUNT_ID}:role/GithubActionsDeploymentRole
-  memorySize: 1024
+  memorySize: 256
```
FergusMcGlynn, line 10: Here is a graph showing the cold start time vs the warm start time for Course Search with different Lambda memory settings. There is a noticeable cold start penalty for 256 Mb - but it's not completely obvious what the optimal amount is (given that there's more memory costs more to run). My inclination is to stick with 1024 Mb memory as it offers a ~1s improvement on 256 Mb, but I'd like to hear other opinions. 

image

BTW it might be worth repeating this experiment if/when we introduce CloudFront.

FergusMcGlynn (changes requested):

serverless.yml
ChristianBeddows: Is there any way that we can distinguish cold starts apart from an increased response time? It could be helpful to know what percentage of production executions are cold starts in the future.

FergusMcGlynn: Good question! I don't think AWS Lambda outputs anything (e.g. in its CloudWatch logs) to say whether a particular execution had a cold or a warm start. I suspect we might be able to deduce that an execution was a cold start based on how long it was since the previous execution (e.g. > 15mins since last execution = _probably_ cold start). We won't have any useful data on Course Search usage until it goes live, but we might be able to predict usage by looking at how often Course Finder is used, and whether most searches there occur within 15 mins of each other.

ce666-york: Could we pull out some stats about activity on the current Course Finder from Google Analytics, and do a brief cost analysis of the lambda for different memory values? That data could support this decision 📈

As an aside, I think we'll need to implement some sort of health check for Course Search, so a probable side-effect of that will be keeping the lambda warm even if activity does fall off. That doesn't account for multiple searches occurring at the same time though.

FergusMcGlynn: I've spent a little time this afternoon trying to calculate Lambda costs for a couple of scenarios. My conclusion is that, as things stand, we probably don't need to worry about cost much - we're looking at a few cents per month no matter what we set the Lambda memory to.

How much Lambda execution time does a single course search currently take (average of 5 page loads)?

With 256 Mb cold start: **3,360 ms**
With 256 Mb warm start: **1,800 ms**
With 1024 Mb cold start: **1,640 ms**
With 1024 Mb warm start: **1,420 ms**
With 3008 Mb cold start: **1,480 ms**
With 3008 Mb warm start: **1,360 ms**

Looking at Google Analytics for Course Finder, we had 5,664 undergraduate course searches per month on average over the last 12 months (see end for calculation)

Since we don't know what proportion of course searches on the new system will be cold starts or warm starts, lets calculate best and worst case scenarios for different memory settings to give us a guide. Worst case scenario is every page load is a cold start. Best case scenario is every page is a warm start.

| Memory  | Scenario   | Seconds per month        | Total compute per month          |
|---------|------------|--------------------------|----------------------------------|
| 256 Mb  | worst case | 5,664 x 3.360s = 19,031s | 19031 x 256/1024  =>  4,758 GB-s |
| 256 Mb  | best case  | 5,664 x 1.800s =  9,803s |  9803 x 256/1024  =>  2,451 GB-s |
| 1024 Mb | worst case | 5,664 x 1.640s =  9,289s |  9289 x 1024/1024 =>  9,289 GB-s |
| 1024 Mb | best case  | 5,664 x 1.420s =  8,043s |  8043 x 1024/1024 =>  8,043 GB-s |
| 3008 Mb | worst case | 5,664 x 1.480s =  8,383s |  8383 x 3008/1024 => 24,625 GB-s |
| 3008 Mb | best case  | 5,664 x 1.360s =  7,703s |  7703 x 3008/1024 => 22,628 GB-s |


AWS Lambda in region eu-west-1 costs:
- Requests: $0.20 per 1M requests
- Compute:  $0.0000166667 for every GB-second

**Requests**

Course Search currently makes 10 Lambda requests per search (1 main one, and 9 static assets). So per month there are 56,640 requests. 
56,640 / 1,000,000 = 0.05664 x 0.20 = $0.01 for a month's worth of Course Search requests

Note 1: $0.20 per 1M requests will be shared amongst all ESG Lambda executions - $0.01 just represents the Course Search portion of that. It may be that AWS charge $0.20 if you make < 1M requests and not a fractional proportion of it.

Note 2: AWS give each account 1M free requests per month. Depending on ESG usage as a whole, we may not have to pay anything for these requests.

**Compute**

| Memory  | Scenario   | Compute cost per month                          |
|---------|------------|-------------------------------------------------|
| 256 Mb  | worst case |  4,758 GB-s x $0.0000016667 => $0.0079 => $0.01 |
| 256 Mb  | best case  |  2,451 GB-s x $0.0000016667 => $0.0041 => $0.01 |
| 1024 Mb | worst case |  9,289 GB-s x $0.0000016667 => $0.0155 => $0.02 |
| 1024 Mb | best case  |  8,043 GB-s x $0.0000016667 => $0.0134 => $0.01 |
| 3008 Mb | worst case | 24,625 GB-s x $0.0000016667 => $0.0410 => $0.04 |
| 3008 Mb | best case  | 22,628 GB-s x $0.0000016667 => $0.0377 => $0.04 |


Note: AWS give each account 400,000 GB-s free per month. Course Search looks like it won't use anywhere near that much. Depending on ESG usage as a whole, we may not have to pay anything for compute.

**Health check thought**

Chris mentioned that at some point in the future we might want to introduce a health check for Course Search. This would involve pinging it every, say, 10 minutes to check whether it's up or down, and would have the side effect of keeping the Lambda warm (if executed more frequently than every ~15 mins). If we polled Course Search every 10 mins, that would add an extra 4,380 additional page loads per month, which would (very roughly) double the amount of compute time and therefore cost.

**CloudFront thought**

If we introduce CloudFront and cache static assets, that will change these calculations. The Lambda cost will be lower, because each search will only require one Lambda invocation rather than 10. But there are additional costs for running a CloudFront distribution.

**Appendix** - monthly unique undergraduate searches for Course Finder from Google Analytics:

    October 2020:   7,325
    September 2020: 7,288
    August 2020:    5,630
    July 2020:      5,222
    June 2020:      5,049
    May 2020:       4,288
    April 2020:     4,133
    March 2020:     4,306
    February 2020:  5,176
    January 2020:   6,580
    December 2019:  5,383
    November 2019:  7,592

ce666-york: Thank you for pulling these numbers together, very detailed!

Do you know why [the lambda cost calculator on the page you link](https://aws.amazon.com/lambda/pricing/) comes out at ~$63? This is for a 3GB instance, with 845620 requests at 1480ms, and without the free tier.

I also noticed on that page it said price per 100ms varies depending on the memory allocation, so it's not always $0.0000016667 - though I think you may have factored that in with your varying compute throughput values?

Either way I think given the effect CloudFront will have on things as well as the costs of the other services, there are enough variables that we'll want to do some ongoing cost monitoring and make further decisions based on real numbers (e.g. 3GB may well be very reasonable with CloudFront caching 9/10 requests). For now, even if the numbers were to come out like the cost calculator example above, 1024 it's still very reasonable in my view, so I'm happy to go with your suggestion.

It would be great to document this investigation and the decision we come to... :slightly_smiling_face:

FergusMcGlynn: The discrepancy with $63 on the 3Gb instance: when we do a course search currently, it executes the Lambda 10 times - 1 of which takes ~600ms (the main page), and 9 of which take 100ms (static assets - they actually take 1-2ms but Lambda charges in 100ms blocks). The whole 10 requests takes (on average) 1480ms. Not 1480ms for each of the 10 requests. Hence the calculation is 84,562 x 1420ms rather than 845,620 x 1420ms.

The price as far as I can tell is always $0.000016667 per GB-s in region eu-west-1. The variation in the figures quoted for 100ms are simply that figure divided by 10 (to get 100ms) and then multiplied by the relevant fraction of 1GB (e.g. multiplied by 256/1024, or 1024/1024, or 3008/1024). 

I agree that if we add a CloudFront distribution and cache static assets, i.e. only invoke one Lambda execution per course search rather than 10, the Lambda costs will be significantly lower that what I quoted. Although the 1 Lambda invocation that we will continue to do will be the long one.

For 256Mb cold, that'd take ~1700ms (vs 3360ms for all 10)
For 256Mb warm, that'd take ~900ms (vs 1800ms for all 10)
For 1024Mb cold, that'd take ~800ms (vs 1640ms for all 10)
For 1024Mb warm, that'd take ~600ms (vs 1420ms for all 10)
For 3008Mb cold, that'd take ~600ms (vs 1480ms for all 10)
For 3008Mb warm, that'd take ~500ms (vs 1360ms for all 10)

From the above, it looks like adding CloudFront would (very roughly) halve the Lambda costs - but then add a bunch of CloudFront costs. It might work out more expensive - but hopefully more performant.

ce666-york: Ah the 1480ms for 10 requests was the part I was missing - that all makes sense.

I'm in an odd position where I raised this PR so I can't approve, but just pretend that I have :grin:

FergusMcGlynn:
```suggestion
  memorySize: 1024
```

FergusMcGlynn (approved):

PR #86, opened 2020-11-20
Run visual testing on push to dev

ce666-york:
Two things I noticed when testing Percy:

- We don't take snapshots when pushing to our default branch - future branches won't compare changes against the correct base snapshot unless we do this.
- For whatever reason the Percy script still runs forever on my local machine unless `process.exit()` is present - I saw no harm in adding it so that it will hopefully work for everyone.

### DIFF
--- a/.github/workflows/percy-visual-testing.yml
+++ b/.github/workflows/percy-visual-testing.yml
@@ -2,6 +2,9 @@ name: Run visual tests with Percy
 on:
   pull_request:
   workflow_dispatch:
+  push:
+    branches:
+      - dev
 
 jobs:
   visual-tests:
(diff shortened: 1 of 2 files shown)


mrgibbonsuoy:
> * We don't take snapshots when pushing to our default branch - future branches won't compare changes against the correct base snapshot unless we do this.

Ok: I left it out as we'd be pushing into Dev and checking there. However I'll add it in and put a PR in.

RichardAllum (approved):

ce666-york:
> Ok: I left it out as we'd be pushing into Dev and checking there. However I'll add it in and put a PR in.

`dev` is our default branch. All the changes I'm describing are part of this PR - worth reading [the documentation](https://docs.percy.io/docs/baseline-picking-logic) if something's not clear

mrgibbonsuoy (approved):

ncn1 (approved):

FergusMcGlynn (approved):

PR #87, opened 2020-11-23
CFIN-390 Local performance testing scripts

ChristianBeddows:
Please check this out locally and have a crack at using the npm scripts. You'll need to be running a production server on port 3000 prior to the run for this to work.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "private": true,
   "scripts": {
     "dev": "next dev",
+    "prod": "npm run build && npm run start",
     "build": "next build",
     "start": "next start",
     "package": "next build && sls package",
@@ -17,7 +18,8 @@
     "check": "npm run checkformat && npm run lint && npm run test",
     "formatandcheck": "npm run format && npm run lint && npm run test",
     "fc": "npm run formatandcheck",
-    "percy": "percy exec -- node src/tests/visual/percyVisualTesting.js"
+    "percy": "percy exec -- node src/tests/visual/percyVisualTesting.js",
+    "lighthouse": "lighthouse --output-path=./lighthouse.html --chrome-flags=\"--headless\" http://localhost:3000/?search=maths --view"
   },
   "dependencies": {
     "@university-of-york/esg-lib-pattern-library-react-components": "^6.1.3",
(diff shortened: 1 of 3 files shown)


ncn1:
> You'll need to be running a production server on port 3000 prior to the run for this to work.

ie. run the 'prod' npm script

FergusMcGlynn (approved):
I've run this locally and it seems to work fine for both desktop and mobile. I get really bad scores for desktop, but that's another ticket.

ncn1 (approved):

package.json
```diff
@@ -4,6 +4,7 @@
   "private": true,
   "scripts": {
     "dev": "next dev",
+    "prod": "npm run build && npm run start",
```
ce666-york, line 7: Does this use development or production environment variables? Wondering if it's slightly misnamed and better to just roll into `start`

FergusMcGlynn: It'll be running production environment variables

ce666-york: Thanks Ferg. I think in my head there's a slightly separate (and wider) issue of local environment variables, when you might expect to use them, and what our defaults are. This also ties into the question of ideally how would we want to develop locally, as currently there's a dependency on the development or production API.

ce666-york (changes requested):
These scripts seem to give significantly worse performance scores for both desktop (mid 60's) and mobile (mid 30's) compared to what running Lighthouse in Chrome/Pagespeed insights give (mid 90's for desktop and mid 60's for mobile).

I think we need to work out why this is the case before we can use these scripts - looks great so far but I think there's a bit more to do!

ce666-york (approved):
(Should have marked as approved as in principle I think these changes are fine to be merged!)

ce666-york (changes requested):
Actually I think this PR should be for merging into the story branch `CFIN-140` rather than `dev`. It'll probably be good to merge `CFIN-140` back into your branch too so you can develop against the build with the performance improvements.

mrgibbonsuoy (approved):

ChristianBeddows:
I'm closing this PR as I need to do some git tidying, and further changes to get the script working at an acceptable level.

PR #90, opened 2020-11-26
CFIN-390 Local Lighthouse reporting

ChristianBeddows:
I've updated the lighthouse npm tasks so that they now reference exactly the same configuration as the automated performance tests, which seems to resolve the low scores we were getting previously. This configuration comes straight from the Lighthouse npm module - so we won't need to maintain it if Google release any updates. 

**Can reviewers please check out this branch, run the reports and check that the score is appropriate**


package.json
```diff
@@ -17,7 +18,10 @@
     "check": "npm run checkformat && npm run lint && npm run test",
     "formatandcheck": "npm run format && npm run lint && npm run test",
     "fc": "npm run formatandcheck",
-    "percy": "percy exec -- node src/tests/visual/percyVisualTesting.js"
+    "percy": "percy exec -- node src/tests/visual/percyVisualTesting.js",
+    "lighthouse:desktop": "lighthouse --output-path=./lighthouse.desktop.html --config-path=./node_modules/lighthouse/lighthouse-core/config/lr-desktop-config.js --chrome-flags=\"--headless\" http://localhost:3000/?search=maths --view",
+    "lighthouse:mobile": "lighthouse --output-path=./lighthouse.mobile.html --config-path=./node_modules/lighthouse/lighthouse-core/config/lr-mobile-config.js --chrome-flags=\"--headless\" http://localhost:3000/?search=maths --view",
```
ChristianBeddows, line 23: The config path pointing towards `node_modules` is doing the same thing as `mobileConfig = require(<path>);` in the automated script

ce666-york (approved):
This works for me, and the numbers from the generated report, the automated check, and running Chrome devtools locally all seem to broadly line up.

The numbers are noticeably better than what Chrome/Pagespeed Insights give us when pointing to dev/production - I guess this is inevitable given we're testing locally, but it makes me think we should bump the local mobile threshold up to 90.

We could also investigate an automated check to monitor performance in the live application in future.

ncn1:
> Can review



> Can reviewers please check out this branch, run the reports and check that the score is appropriate

I get 71 for desktop and 21 for mobile - presumably not appropriate?

mrgibbonsuoy:
I received 65 on Desktop and 13! for mobile. I tried to run again and lighthouse will not render a report. So whilst I'm providing my results, I'd not trust them. May be a reboot post our meetings and try again.

ChristianBeddows:
I'm going to merge this into 140 so it doesn't go stale and get forgotten. 

The score issues seem to vary between machines, so I think some further investigation might be warranted in the future if this is a persistent issue.

PR #93, opened 2021-01-06
(Story merge) CFIN-140 Performance enhancements

ChristianBeddows:
This is a story level merge of the entirety of CFIN-140. I believe all parts of this has been individually reviewed, but as it isn't the biggest changeset in the world I'd recommend skimming through anyway.

Includes:
 - Lighthouse testing
 - GitHub performance-check workflow
 - A couple of performance enhancments


ChristianBeddows:
Could someone else take a look at the percy diff? Looks like the cookie banner has disappeared, but I'm not sure why. Footer looks shorter, but the content seems the same

ce666-york (approved):

FergusMcGlynn (approved):

PR #96, opened 2021-01-19
CFIN-402 Create new rich tags

ncn1:


src/components/RichTag.js
```diff
@@ -0,0 +1,48 @@
...
+};
+
+const ArrayTag = ({ title, content }) => {
+    const contents = content.map((item) => (
+        <ArrayTagContent key={item.text} icon={item.icon} label={item.text} fullWidth={title === null} />
+    ));
+
+    return (
+        <li className="rich-tag">
+            {title && <dt className="rich-tag__title">{title}</dt>}
```
FergusMcGlynn, line 25: A `<dt>` is [supposed to sit within a description list](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dt) (`<dl>` tag) - is `<dt>` the appropriate tag for this situation?

src/tests/components/CourseDetails.test.js
```diff
@@ -3,7 +3,8 @@ import React from "react";
 import { CourseDetails } from "../../components/CourseDetails";
 
 describe("Course details", () => {
-    it("works with no metadata", () => {
+    /* Commenting out the following test until the hard-coded extras tags are removed */
```
FergusMcGlynn, line 6: You can ignore tests by adding `.skip` to the keyword `it`, `test` or `describe`, e.g. 
```
it.skip("works with no metadata", () => {
```
But I would prefer it if we didn't comment out or ignore any tests. Is there any particular reason why `the hard-coded extras tags are removed` can't be done as part of this work?

src/components/RichTag.js
```diff
@@ -0,0 +1,48 @@
...
+    </li>
+);
+
+RichTag.propTypes = {
+    topIcon: PropTypes.string,
+    title: PropTypes.string,
+    subText: PropTypes.string,
+};
+
+const ArrayTag = ({ title, content }) => {
```
FergusMcGlynn, line 18: I can't see anywhere that this function is being used. Is it needed?

src/pages/_app.js
```diff
@@ -0,0 +1,14 @@
...
+const MyApp = ({ Component, pageProps }) => {
+    return <Component {...pageProps} />;
+};
+
+MyApp.propTypes = {
+    Component: PropTypes.any,
+    pageProps: PropTypes.any,
+};
+
+export default MyApp;
```
FergusMcGlynn, line 14: I think this came up in the last PR for this code. `MyApp` doesn't seem like a production-grade function name to me. I also don't really understand what it's doing. I think last time you said that it had been copied and pasted from a website to make something work. It'd be good to see if we can get an understanding of how/why it works, as we'll be maintaining it.

src/components/RichTag.js
ncn1: Part of the problem with this being designed for moving to the pattern library. ArrayTag is for the extras tag, which had demo versions and these have been removed. I don't think we are planning to use the extras tags in the first release. But they will be needed somewhere eventually.

FergusMcGlynn: If we're not going to use `ArrayTag` and `ArrayTagContents` in the first release then I think we should remove the code.

src/components/CourseDetails.js
```diff
@@ -1,57 +1,46 @@
...
 
 const CourseDetails = ({ course }) => {
     return (
-        <TagList>
-            <AwardTag award={course.award} />
-            <StartDateTag yearOfEntry={course.yearOfEntry} />
-            <LengthCourseTag length={course.length} />
-        </TagList>
+        <RichTagList>
+            <LevelAndAwardTag level={course.level} award={course.award} />
```
FergusMcGlynn, line 10: Do we need to tidy up the case of `course.level`? By default, I'm guessing this will output "undergraduate" for all undergraduate courses - but in the designs, it states "Undergraduate".

src/tests/components/CourseDetails.test.js
```diff
@@ -12,43 +13,112 @@ describe("Course details", () => {
...
 
-        expect(tag).toHaveTextContent("BA (Hons)");
-        expect(within(tag).getByTestId("tag-icon")).toHaveClass("c-icon--university");
+        expect(tag[0]).toHaveTextContent("undergraduate");
+        expect(tag[0]).toHaveTextContent("BA (Hons)");
+        expect(within(tag[0]).getByTestId("tag-icon")).toHaveClass("c-icon--mortar-board");
     });
 
-    it("displays start date tag", () => {
+    it("displays level and award tag when both only level is present", () => {
```
FergusMcGlynn, line 33: Test name needs a little tidying
```suggestion
    it("displays level and award tag when only level is present", () => {
```

src/components/CourseDetails.js
ncn1: The css takes care of it.

src/tests/components/CourseDetails.test.js
```diff
@@ -12,43 +13,112 @@ describe("Course details", () => {
...
 
-        expect(tag).toHaveTextContent("4 years full-time");
-        expect(within(tag).getByTestId("tag-icon")).toHaveClass("c-icon--clock-o");
+        if (tag.length === 0) {
+            const tagList = screen.getByRole("list");
+            expect(tagList).toBeEmptyDOMElement();
+        } else {
+            expect(tag[0]).toHaveTextContent("year abroad");
+            expect(within(tag[0]).queryByTestId("tag-icon")).toBe(null);
+        }
```
FergusMcGlynn, line 122: I'm confused by this `if`/`else` - it's a bit odd to see a conditional in a test, and I'm not really sure what it's achieving here.

src/tests/components/CourseSearchResults.test.js
```diff
@@ -10,8 +10,10 @@ describe("CourseSearchResults", () => {
 
         render(<CourseSearchResults isSuccessfulSearch searchResults={searchResults} searchTerm="foobar" />);
 
-        expect(screen.getByRole("link", { name: "Maths" })).toBeInTheDocument();
-        expect(screen.getByRole("link", { name: "Physics" })).toBeInTheDocument();
+        const headings = screen.getAllByRole("heading");
+
+        expect(headings[0]).toHaveTextContent("Maths");
+        expect(headings[1]).toHaveTextContent("Physics");
```
FergusMcGlynn, line 16: Any particular reason for switching from `link` role to `heading` role for these tests?

src/tests/components/RichTagList.test.js
```diff
@@ -0,0 +1,71 @@
...
+        expect(within(tags[0]).getByTestId("tag-icon")).toHaveClass("c-icon--institution");
+
+        expect(tags[1]).toHaveTextContent("fruit salad");
+        expect(within(tags[1]).getByTestId("tag-icon")).toHaveClass("c-icon--key");
+
+        expect(tags[2]).toHaveTextContent("hot pudding");
+        expect(within(tags[2]).getByTestId("tag-icon")).toHaveClass("c-icon--plane");
+    });
+
+    it("displays extra tags as per inputs", () => {
```
FergusMcGlynn, line 29: I think we can delete this test, or move it to a branch where 'extras' tag work can be stashed, if the 'extras' tag isn't required in the first release.

src/tests/pages/index.test.js
```diff
@@ -37,8 +37,10 @@ describe("App", () => {
     it("displays the titles from course search results", () => {
         render(<App isSuccessfulSearch searchResults={searchResults} searchTerm="foobar" />);
 
-        expect(screen.getByRole("link", { name: "Maths" })).toBeInTheDocument();
-        expect(screen.getByRole("link", { name: "Physics" })).toBeInTheDocument();
+        const headings = screen.getAllByRole("heading");
+
+        expect(headings[2]).toHaveTextContent("Maths");
+        expect(headings[3]).toHaveTextContent("Physics");
```
FergusMcGlynn, line 43: Similar to another file - is there any particular reason to select by the `heading` role rather than the `link` role here? Referring to `headings[2]` and `headings[3]` isn't particularly intuitive to a newcomer.

styles.css
```diff
@@ -0,0 +1,38 @@
...
+    flex-direction: row;
+    flex-wrap: wrap;
+    margin: 0 0 20px;
+    padding: 0;
+}
+
+.rich-tag {
+    box-sizing: border-box;
+    background-color: #f0f2e9;
+    text-transform: capitalize;
```
FergusMcGlynn, line 12: This should be fine for the tag content that we want to display in the first release. We might need to change it in future though - a tag for "Year in Industry" would be transformed to "Year In Industry" by this. Also "Available for clearing" would become "Available For Clearing". Do we want to cross that bridge when we come to it, or deal with it now?

FergusMcGlynn (changes requested):
There are a couple of issues where I've commented that I think it'd be good to resolve before merging.

src/components/RichTag.js
RichardAllum: Good spot, this should be either a `<div>`, a `<p>` or a `<span>` - this will be left in from where we tried using a description list but found it didn't work with icons

src/tests/components/CourseDetails.test.js
ncn1: They have been removed - I will reinstate this test

src/tests/components/CourseDetails.test.js
ncn1: Another leftover from the demo tags

src/tests/pages/index.test.js
ncn1: It's a more specific test - these items are headings and come in this order.

styles.css
ncn1: I think it looks OK https://uoy.enterprise.slack.com/files/W013GPC4T7G/F01J7NLTY22/image.png, 
so only change it later if someone specifically notices and objects to "Year In Industry"

src/tests/components/CourseSearchResults.test.js
ncn1: Allows for the testing of the order they appear in as well as the fact they do appear.

styles.css
FergusMcGlynn: I think "Year In Industry" and "Available For Clearing" might not conform to the [University style guide](https://www.york.ac.uk/staff/external-relations/print-marketing/style/). Under "capitalisation" it says
```
Do not capitalise:
    * the smaller words in titles such as in, at, of, the, and, on
````
It also links to the [digital content style guide](https://wiki.york.ac.uk/display/DM/Digital+content+style+guide), which has a note about [capitalisation of headings](https://wiki.york.ac.uk/pages/viewpage.action?pageId=74712085):
```
All headings and navigation should be written in sentence case
```
I realise these are not headings - so maybe the rule doesn't apply here. I think it might be worth querying. It would be a bit odd to forbid (for example) "4 Years Full-Time" in a heading, but permit it in a rich-tag/key-fact. I'll post something in Slack and see what others think.

File: src/tests/pages/index.test.js (hunk 10 above)
Comment (FergusMcGlynn) [reply src/tests/pages/index.test.js:43]: The specific ordering improvement comes from the `getAllBy...` (instead of `getBy...`), not because we've switched from selecting links to selecting headings. From a user's perspective, I'd have thought it's more important that there's a link to a course page (with descriptive link text) than that there's a heading with a course's title in it.

I've also just found, in the [University's digital style guide](https://wiki.york.ac.uk/pages/viewpage.action?pageId=74712085#Style(contentpatterns)-Headingsandsub-headings), the following:
```
Don't put links in headings
```
This is news to me, and I'll post something in Slack for the team to discuss - I don't think we need to resolve that in this PR. It might not be relevant because, technically, we're putting a heading in a link, but there might be a style principle here that we need to consider. It makes me even more wary of testing by heading rather than testing by link

Comment (FergusMcGlynn) [reply src/tests/pages/index.test.js:43]: Also, regarding the point about `headings[2]` and `headings[3]`, it's a bit confusing for a reader - where are `headings[0]` and `headings[1]`? Why is `2` a magic number? I'm guessing that `headings[0]` refers to the `<h1>` page title and `headings[1]` refers to... some other heading somewhere, but it's not intuitive. The reader needs to know some things that are outside the test in order to understand the test.

File: styles.css (hunk 11 above)
Comment (FergusMcGlynn) [reply styles.css:12]: From what Paul Kelly has posted in Slack, it looks like we should be aiming for `4 years full-time` rather than `4 Years Full-Time`. It doesn't look like there's a CSS `text-transform` value for sentence case, but I found [a potential workaround](https://css-tricks.com/almanac/properties/t/text-transform/#comment-1607495) online. It might need some tweaking, but maybe worth a try?

Comment (ncn1) [reply styles.css:12]: Good find!

File: styles.css [hunk 12]
```diff
@@ -0,0 +1,45 @@
...
+.rich-tag__text {
+    font-size: 14px;
+    font-size: 0.875rem;
+    margin: 0;
+    line-height: 1.45;
+}
+
+.rich-tag__text::first-letter {
+    text-transform: capitalize;
+}
```
FergusMcGlynn, line 45: Is there any way we can test this? It'd be nice to be able to verify that this does what we're intending.

ncn1: I think only by viewing the Percy snapshot. In CourseDetails.test.js, it was passing this test "displays tag with start year and length of course when both are present" which contains `expect(tag[0]).toHaveTextContent("4 years full-time");` when it was actually showing as '4 Years Full-Time'

FergusMcGlynn: Maybe something like Cypress could test this? We've been talking for a while about investigating Cypress testing (which I believe would be akin to Geb tests for a Grails application). This might a be a good opportunity to see whether that's a worthwhile option, if there's no way of testing CSS changes using React Testing Library?

ChristianBeddows (approved):

styles.css
ChristianBeddows: Interesting - `toHaveTextContent` should be case sensitive

Screenshot from 2021-01-25 11-21-29

ChristianBeddows: Although thinking about it - if the CSS is doing the capitalising, then it's text content was always `4 years full-time` - which might explain why it was passing! Would it be possible to use a `.toBeDisplayed` method to ensure it's viewable as we expect?

ChristianBeddows: I notice there's a method called `getByDisplayValue` - perhaps this would take the CSS capitalisation into account?

Screenshot from 2021-01-25 11-28-49

FergusMcGlynn: Richard's right that text-transform [doesn't change the DOM](https://github.com/testing-library/react-testing-library/issues/567). I disagree though that we'd just be testing standard CSS - we'd be testing whether or not we're using it correctly to achieve what we want to achieve.

ncn1: I think that investigating Cypress testing is beyond the scope of this story and PR, (which is beginning to seem never ending!) The Percy test has caught the fact that the page did change as a result of this css change, and I in fact thought that was the point of adding Percy, to detect changes made by css that not easily testable with the React testing library.

RichardAllum: I think Cypress testing may be worth adding to this project at some point, but I also don't think this PR is the place to start - the text capitalisation is not in the acceptance criteria, and adding a new testing framework to the project for it at the final hurdle feels like overkill to me. I'd be happy to put a story in the backlog for Cypress testing on a more generic level though, as there would be user benefit to it.

FergusMcGlynn: I see Percy as being more a safety check to ensure that a code change hasn't introduced an unintended UI change. I wouldn't even call Percy a "test" - it's a check, it can't validate that the application delivers on its requirements.

We have a requirement - "text in tags should be presented in sentence case". We have input data, from the Courses API. We can write a test that verifies whether, no matter what the input data is, the output is presented in sentence case. This is testable and, if not crazy amounts of effort, worth testing, as are all requirements of the application.

It may be that it would take crazy amounts of effort to do such a test, in which case we might decide that we don't have the resources to do it. But I don't think it's quite right to say that Percy is capable of validating this requirement.

ce666-york: I also see visual testing as an aid or a prompt rather than a substitute for functional tests.

I've been wanting us to look into Cypress for a while, I don't mind being flexible and including it in this story, but if it's a lot of work happy to split it out too. Reading the docs, it does sound interesting and quite straightforward https://docs.cypress.io/guides/getting-started/writing-your-first-test.html.

FergusMcGlynn: Agreed that "the text capitalisation is not in the acceptance criteria" - but that's an argument for not implementing a text capitalisation solution. I don't see it as an argument for not testing a text capitalisation solution. Too many "not"s - but I hope you see what I mean.

PR #97, opened 2021-01-26
CFIN-371 Inclusion of rich tags, without any text transformation

ncn1:
Mostly all reviewed in this closed pull request https://github.com/university-of-york/uoy-app-course-search/pull/96, except text-transform lines missing from styles.css

### DIFF
--- a/src/tests/pages/index.test.js
+++ b/src/tests/pages/index.test.js
@@ -37,10 +37,8 @@ describe("App", () => {
     it("displays the titles from course search results", () => {
         render(<App isSuccessfulSearch searchResults={searchResults} searchTerm="foobar" />);
 
-        const headings = screen.getAllByRole("heading");
-
-        expect(headings[2]).toHaveTextContent("Maths");
-        expect(headings[3]).toHaveTextContent("Physics");
+        expect(screen.getByRole("heading", { name: "Maths" })).toBeInTheDocument();
+        expect(screen.getByRole("heading", { name: "Physics" })).toBeInTheDocument();
     });
 });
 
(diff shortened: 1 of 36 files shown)


src/tests/pages/index.test.js
```diff
@@ -37,8 +37,10 @@ describe("App", () => {
     it("displays the titles from course search results", () => {
         render(<App isSuccessfulSearch searchResults={searchResults} searchTerm="foobar" />);
 
-        expect(screen.getByRole("link", { name: "Maths" })).toBeInTheDocument();
-        expect(screen.getByRole("link", { name: "Physics" })).toBeInTheDocument();
+        const headings = screen.getAllByRole("heading");
+
+        expect(headings[2]).toHaveTextContent("Maths");
+        expect(headings[3]).toHaveTextContent("Physics");
```
ce666-york, line 43: I believe there was an unresolved discussion about magic numbers in the previous PR. I think the risk here is that if we add new components and/or headings before the rendering of results, this test will fail - for me ideally that wouldn't be the case, as we would still be meeting the acceptance criteria in the test's title.

I think `CourseSearchResults.test.js` is the right place to test the ordering (which you've done!), I'd suggest keeping it more generic here, which also removes that prerequisite knowledge that there are (currently) two headings before the results:

```suggestion
        expect(screen.getByRole("heading", { name: "Maths" })).toBeInTheDocument();
        expect(screen.getByRole("heading", { name: "Physics" })).toBeInTheDocument();
```

I don't think it's a huge deal, but it's nice to avoid issues in future if we spot them now :slightly_smiling_face:

ce666-york (approved):

ce666-york:
I've approved, but think it's worth @FergusMcGlynn reviewing before merging!

FergusMcGlynn (approved):
I agree with Chris that it'd be nice if the test that currently references `headings[2]` and `headings[3]` could be made a little more robust - I think the suggestion there looks good - although it's not a huge deal.

RichardAllum (approved):

ChristianBeddows (approved):

PR #99, opened 2021-01-27
CFIN-446 Create HeroBanner component

ChristianBeddows:


RichardAllum:
This is a draft PR to view the changes in Percy/let others view them

styles.css
```diff
@@ -35,3 +35,16 @@
     margin: 0;
     line-height: 1.45;
 }
+
+.c-hero__banner {
+    background: linear-gradient(180deg, #138BB3 7.81%, #0F7799 100%);
```
ncn1, line 40: When I ran and viewed this locally, I could not visually see anything different when I changed the css to just be 'background: #0F7799;'. Is it meant to be very subtle?

PR #100, opened 2021-01-28
CFIN-429 Add brief info about the rich tags into the README

ncn1:


README.md
```diff
@@ -57,8 +57,12 @@ npm ERR! 404 Not Found - GET https://registry.npmjs.org/@university-of-york%2fes
 npm ERR! 404
 npm ERR! 404  '@university-of-york/esg-lib-pattern-library-react-components@4.3.4' is not in the npm registry.
 ```
-#### A pages/_app.js file exists so an internal css file can be used
-See the [Next Built-In CSS Support - Adding a Global Stylesheet](https://nextjs.org/docs/basic-features/built-in-css-support) documentation
+#### Use of local 'styles.css' file
+The pages/_app.js file exists so this internal css file can be used - 
+see the [Next Built-In CSS Support - Adding a Global Stylesheet](https://nextjs.org/docs/basic-features/built-in-css-support) documentation.
+The rich tags are currently being defined in this file as part of the application, but may be moved to the React Pattern library in future, 
+depending on the outcome of discussions with the Digital Platforms team.
```
FergusMcGlynn, line 64: I'm not sure we need the last sentence. It could be helpful for someone wanting to know where the rich tags CSS is defined, but I think that can probably be deduced by a developer from looking at the code. Not sure the README should include speculation about future plans - not that it's harmful, just not essential. I appreciate that it's not obvious what should go in and what shouldn't, but I think the general principle is it should contain information a developer needs to get up and running (and anything else should be somewhere else, e.g. wiki).

I also think the first sentence could be clarified slightly:
```suggestion
#### Use of local 'styles.css' file
The `pages/_app.js` script enables an internal css file (`styles.css` in the root folder) to be used - 
see the [Next Built-In CSS Support - Adding a Global Stylesheet](https://nextjs.org/docs/basic-features/built-in-css-support) documentation.
```

FergusMcGlynn:
Unrelated to this PR - I only just noticed that `styles.css` is in the project's root folder. Apologies for not picking up on it in previous PRs. I'm not sure that `styles.css` should be there. The root folder is (or has become) a place for configuration files. I think `styles.css` would be more appropriate in the `src` or `src/pages` folder. Is there any particular reason why it needs to be in the root folder? Is it a Next.js convention? What do others think?

README.md
ncn1: In that case, PR is no longer about the quoted JIRA story, so I should probably just close it. I'll move your other comment into Slack for discussion, so it doesn't get lost.

PR #101, opened 2021-01-28
CFIN-433 Add rudimentary css to course-search app

RichardAllum:
This commit will provide us with the core user benefit from CFIN-433, while work is done to get the styles into the pattern library. This way if there is pushback about the styles being in pattern libraries we will still have the benefit


ncn1:
I've checked out this branch and run it locally, nd it looks identical to the staging version, white space below the footer.

PR #102, opened 2021-01-28
BUGFIX - add media queries so tags adapt to mobile

RichardAllum:
I didn't catch this on the initial review, but when migrating to the pattern library some media queries were added to allow the tags to reflow at mobile and tablet views to better match the preview.


ncn1:
What's the best way to check this? Running this and shrinking my browser window as small as it will go behaves in the same way for this version and the staging version.

mrgibbonsuoy (approved):

RichardAllum:
> What's the best way to check this? Running this and shrinking my browser window as small as it will go behaves in the same way for this version and the staging version.

The effect should be minimal, but running it with a thin browser window should look like this
Screenshot from 2021-01-29 09-24-38
instead of having the tags on top of each other.

ncn1:
> instead of having the tags on top of each other.
But I don't see the tags on top of one another in the staging version when I shrink the window - do you? If not, this change isn't required.

ChristianBeddows:
> What's the best way to check this? Running this and shrinking my browser window as small as it will go behaves in the same way for this version and the staging version.

@ncn1 Chrome has the device emulator devtool which makes it really easy to test the responsiveness of the page as it resizes. If you set the device to `responsive` at the top you can resize the page to any dimensions by dragging!

Screenshot from 2021-02-01 10-52-08

ChristianBeddows (approved):

ncn1:
Thanks for the reminder - both this version and the staging version show in this tool exactly the same for me, with the tags stacked one on top of the other - I'm still not convinced this change makes any difference.

PR #103, opened 2021-01-28
CFIN-379 Make course titles weightier

ncn1:
The prototype shows the Course titles as 
` font weight: 700, font-size: 24px`
Before this change, our version had
`font weight: 400, font-size: 24px `

The html for the course title is
`<h2 className="c-listings-item__title">{title(course)}</h2>`

\<h2> in the university stylesheet has a font weight of bold, but the c-listings-item__title class font weight of 400 was winning out. So the prototype already conforms to the university styles, just the implementation not working.

The solution has been to create a specific `h2.c-listings-item__title` entry in the styles.css.

### DIFF
--- a/styles.css
+++ b/styles.css
@@ -35,3 +35,7 @@
     margin: 0;
     line-height: 1.45;
 }
+
+h2.c-listings-item__title {
+    font-weight: bold;
+}


RichardAllum (approved):

styles.css
```diff
@@ -35,3 +35,7 @@
     margin: 0;
     line-height: 1.45;
 }
+
+h2.c-listings-item__title {
+    font-weight: bold;
+}
```
FergusMcGlynn, line 41: Are there any classes in the pattern library that we make use of (in addition to the `c-listings-item__title` class already being used) to achieve this? It'd be nice to not have to add our own styles if possible, so that we don't have to worry about keeping in sync with any changes to the pattern library.

ChristianBeddows (approved):
Good PR description!

ce666-york:
What's happening with this PR? Is this the one we're waiting for Digital Platforms to make pattern library changes for, and if so should we just close this?

ncn1:
> What's happening with this PR? Is this the one we're waiting for Digital Platforms to make pattern library changes for, and if so should we just close this?

Yes, so I will close.

PR #104, opened 2021-01-29
CFIN-441 Add flex wrapper to app to make footer sticky

RichardAllum:
Sorry about the previous review, I tried to use % instead of vh on app-wrapper. Unfortunately that is much more fiddly to use/less generic - this one is using the standard css

### DIFF
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -18,32 +18,35 @@ import { searchForCourses } from "../utils/searchForCourses";
 const App = ({ isSuccessfulSearch, searchResults, numberOfMatches, searchTerm }) => {
     return (
         <>
-            <PageHead search={searchTerm} />
-            <UniversityHeaderWithSearch />
-            <UniversityTitleBar title="Courses" />
+            <div className="app-wrapper">
+                <div className="app-body">
+                    <PageHead search={searchTerm} />
+                    <UniversityHeaderWithSearch />
+                    <UniversityTitleBar title="Courses" />
 
-            <WrappedMainGrid>
-                <GridRow>
-                    <GridBoxFull>
-                        <Search
-                            searchTerm={searchTerm}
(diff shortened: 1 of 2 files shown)


src/pages/index.js
```diff
@@ -18,32 +18,35 @@ import { searchForCourses } from "../utils/searchForCourses";
...
     return (
         <>
-            <PageHead search={searchTerm} />
-            <UniversityHeaderWithSearch />
-            <UniversityTitleBar title="Courses" />
+            <div className="app-wrapper">
+                <div className="app-body">
+                    <PageHead search={searchTerm} />
+                    <UniversityHeaderWithSearch />
+                    <UniversityTitleBar title="Courses" />
```
FergusMcGlynn, line 25: How would you feel about moving these new `div`s and the university header and footer into a `UniversityWrapper` component like in the [Directory](https://bitbucket.org/university-of-york/esg-app-directory-frontend/src/master/src/components/UniversityWrapper/UniversityWrapper.js)? Just a thought. 

Because this is the main page for the whole application, and probably the entry point for a new developer trying to understand the application, it'd be nice to keep it as succinct as possible, and introducing a couple of styling `div`s makes it just slightly less easy to read and understand. A `UniversityWrapper` type component would help make it clear that the styling `div`s are part of the header/footer presentation side of things. But as I say, it's not a huge deal.

ncn1 (approved):

src/pages/index.js
RichardAllum: Sure, I was thinking that would come when we migrated to the react pattern library but I'm fine doing it now

ChristianBeddows (approved):

ce666-york:
Has this change been superseded now it's in the react pattern library?

mrgibbonsuoy (approved):

PR #106, opened 2021-02-03
CFIN-372 Add Hero Banner with heading and search box

ChristianBeddows:


src/tests/components/HeroBanner.test.js
```diff
@@ -0,0 +1,14 @@
...
+
+describe("HeroBanner", () => {
+    it("wraps content inside a hero banner", () => {
+        const children = <h2>Hi there</h2>;
+
+        render(<HeroBanner>{children}</HeroBanner>);
+
+        const container = document.querySelector(".c-hero__banner");
+
+        expect(getByText(container, "Hi there")).toBeInTheDocument();
```
ncn1, line 12: Could perhaps use getByRole instead, to 'bake in' accessibility?

ChristianBeddows: That's a good perspective to have when writing tests, but I'm not sure how helpful it would be here. We could get by role `heading` and that would integrate accessibility roles with the test, but this isn't testing an element that's part of our web page/source code - just the element that we've added in this test for the purpose of the test.

That is to say, if the search heading/title on the main page were to change to be less accessible/have a different role, this test wouldn't pick it up.

For that reason, I think it's okay as it is. As this test is searching for text within the container we've created - rather than using magic number indexes - I think it's fine as is and won't break.

ncn1: Now you point it out, I agree that it's not adding anything helpful here - just might be good to get into the habit of using getByRole unless it really can't be used.

ce666-york: I agree with Nicola in that if we can use `getByRole` let's get in the habit of it :slightly_smiling_face:

src/tests/components/HeroBanner.test.js
```diff
@@ -0,0 +1,14 @@
+import { getByText, render } from "@testing-library/react";
+import { HeroBanner } from "../../components/HeroBanner";
+
+describe("HeroBanner", () => {
+    it("wraps content inside a hero banner", () => {
+        const children = <h2>Hi there</h2>;
+
+        render(<HeroBanner>{children}</HeroBanner>);
+
+        const container = document.querySelector(".c-hero__banner");
```
ce666-york, line 10: Should the hero banner have a role?

I think ideally we'd use `getByRole` to find the hero banner rather than `querySelector`, and then use `within` to select components within it - similar to the [tag tests](https://github.com/university-of-york/uoy-app-course-search/blob/dev/src/tests/components/CourseDetails.test.js) (although they had to use `getByTestId` when searching inside the tag, whereas as Nicola pointed out we can use `getByRole`)

FergusMcGlynn: The only [aria role](http://dylanb.github.io/periodic-aria-roles.html) that I think the hero banner could usefully have is `region`. It doesn't fit neatly into any of the [8 landmark roles](https://www.washington.edu/accessibility/web/landmarks/) (in my opinion), and from what I've read, those are the most important ones for users of assistive technology. So this probably isn't a huge deal from an accessibility point of view.

If we think the hero banner is purely presentational, and that a screen reader user would only care about its contents (heading and search), then we don't need to force a role onto it. It's totally fine to let presentational divs be ignored by the accessibility tree, so long as they are genuinely just there for presentation. If we think, though, that it might be useful for a screen reader user to be told that the page's main content comprises a "search form" `region` and a "search results" `region` (for example), then we can add those in using `<section title="search form">` and `<section title="search results">`. I'm not sure if it would make much difference to the page's accessibility (I suspect the user only cares about the actual `search` widget, which is a landmark role, and is already marked up appropriately, within the hero banner) but it probably wouldn't hurt.

I don't really like `document.querySelector(".c-hero__banner");` as a way of selecting an element - it's very much coupled to the implementation, and doesn't read like its testing from the user's perspective. If we were to introduce something like `<section title="search form">` into the hero banner then we should be able to select it using `getByRole("region", { name: "search form"})`. As for verifying its contents, we should be able to do something like

```
expect(screen.getByRole("region", { name: "search form" })).toHaveTextContent("Hi there");
```

or as Chris suggests, maybe something like

```
const hero = screen.getByRole("region", { name: "search form" })
expect(within(hero).getByRole("heading", { name: "Hi there" })).toBeInTheDocument();
```

BTW I don't think "search form" is a particularly great name for that region, if we decided to go down this route. I'm struggling to come up with a better one, which is maybe a sign that it's not particularly useful to denote it as a region in the first place. Just a thought.

FergusMcGlynn: If we decide that the hero banner is purely presentational, then an option could be to give it the role `presentation`, which should cause it to not appear in the accessibility tree, but at the same time, potentially allow us to test it using `getByRole`. I've not used `presentation` before, but it could be worth considering.

public/stylesheets/styles.css
```diff
@@ -46,3 +46,15 @@
     margin: 0;
     line-height: 1.45;
 }
+
+.c-hero__banner {
+    background: linear-gradient(180deg, #138bb3 7.81%, #0f7799 100%);
+    color: white;
+    margin-left: 25%;
+    margin-right: 25%;
```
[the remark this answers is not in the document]
RichardAllum: The first two styles come directly from Jordan's example, while the margins enforce the centering of the content within the `hero` and enforce responsiveness at any content width

src/components/HeroBanner.js
```diff
@@ -0,0 +1,22 @@
+import PropTypes from "prop-types";
+import React from "react";
+
+const HeroBanner = ({ children }) => (
+    <div className="c-hero">
+        <div className="c-hero__banner">
+            <div className="o-wrapper">
```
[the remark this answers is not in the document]
RichardAllum: `o-wrapper` comes with a max width while the `hero` wants to be full width, so we put the wrapper inside to wrap the content. This is our replacement for the normal hero content, which is positioned absolutely and so wouldn't work for us. We chose o-wrapper because it is responsive and will enforce the same content width as the o-wrapper for the main page content, to match Jordan's example.

src/tests/components/HeroBanner.test.js
RichardAllum: Those are good points Ferg! I'm not sure if we'd want to establish a `region` for accessibility for the hero as it's not something I think matters to the page structure for users, but it definitely matters for us.

Since the final version will have Breadcrumbs as the first children of `.c-hero__banner` I don't think that's the place for a search form section, but it could be a good idea to include one in `index.js` around the `<h1>` and search (with the role `search`?). It would then also be a good idea to make the `CourseSearchResults` component a section.

Looking at the structure after this I think the HeroBanner is purely presentational, which makes sense as the change we've made is purely visual. This test is therefore testing something react-testing-library does not like testing, and isn't adding much more value than a snapshot test.

My honest thoughts are that either this test provides value because it's worth testing the specific implementation of this presentational change, or that it doesn't because the testing framework isn't suited for it and instead we should be using Percy for this. I included the test because we've said any new work needs to be tested, but I would probably be in favour of removing it or replacing it with a snapshot since Percy covers the same ground.

src/components/HeroBanner.js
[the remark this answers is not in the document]
ce666-york: I think `o-wrapper` is a top-level building block so it raised a question to see it inside another component (the only other example of this I could see is when an `o-wrapper` is inside the html `header` element - I'd be happy to be directed to other examples though!).

I was wondering about doing something like [the alt grids](https://www.york.ac.uk/pattern-library/layout/alt-grid.html) - what's your feeling on that? If that's a workable solution it'd reduce the css to colour tweaks (which addresses my other comment about css styles) and mean we're using standard pattern library conventions.

src/tests/components/HeroBanner.test.js
FergusMcGlynn: Thanks Richard. Like you, I'm questioning the value of denoting the hero as a `region`. There is already [a `search` role defined on the search form](https://github.com/university-of-york/uoy-app-course-search/blob/dev/src/components/Search.js#L19-L32), and I don't think we need to do any more signposting about where to find the search box than that. I don't think the `<h1>` should be included in the `search` role - it'd probably just confuse people.

I'm also questioning the value of this test. What is it really testing? The author of the React Testing Library has written an article [Write tests, not too many, mostly integration](https://kentcdodds.com/blog/write-tests), which although not specifically addressing this situation has made me think that there are occasions (like this) when unit tests don't really add much value. This function `HeroBanner` is not applying a logical, testable operation to input data, it is merely wrapping the input in some `div`s. We could test that it does indeed wrap the input in some `div`s, but that goes against the philosophy of the React Testing Library, which is to test things from the user's perspective. A snapshot test would be verifying that the generated HTML is what we expect. I'd probably lean towards that in this situation. Don't impose a role on the hero banner, remove this test, add a snapshot test to verify that the generated html meets our expectations. That's just my opinion though.

ncn1: > Don't impose a role on the hero banner, remove this test, add a snapshot test to verify that the generated html meets our expectations.
My opinion too.

ChristianBeddows: Thanks for all your input guys. I think I was thinking along the same lines - this test isn't really providing much value and there aren't many ways that we could change it to be more helpful. I think there's no need to alter the banner with roles as it isn't a particularly important part of the accessibility tree and there's no point doing that just for the tests.

It seems like we're reaching an agreement to just have a snapshot test for this component, so I'll look at doing that instead - feels like the best way to move forward.

public/stylesheets/styles.css
```diff
@@ -46,3 +46,15 @@
...
+}
+
+.c-hero__banner h1 {
+    /*Styling to pad the title so its not pressed against the title bar. We will likely want to remove this when we add the breadcrumbs.*/
+    padding-top: 10px;
```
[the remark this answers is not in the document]
ChristianBeddows: I've added the breadcrumbs component - we can update this later when we have more idea of the web structure but for now it pads out the title for us

src/components/HeroBanner.js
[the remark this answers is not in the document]
ChristianBeddows: I had a crack at using the alt-row stuff and it did not seem to help. Maybe it would warrant more investigation but it doesn't really seem like a good solution - and I'm not really sure what the problem is with using `o-wrapper` in this manner - I thought this was acceptable use of the wrapper?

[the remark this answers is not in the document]
ce666-york: I'll try and throw together a quick example to explain what I mean (if it works). I'd be interested to know why you think it isn't a good solution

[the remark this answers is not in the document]
ce666-york: So I was thinking of something along these lines:
```
<div className="o-wrapper">
    <div className="o-grid">
        <div className="o-grid__row o-grid__row--althero">
            <div className="o-grid__box o-grid__box--full">
                <Breadcrumbs screenBreadcrumbs={[{ label: "Course search", link: "/" }]} />
                <h1>Search undergraduate courses</h1>
                <Search
                    searchTerm={searchTerm}
                    numberOfResultsShown={searchResults?.length}
                />
            </div>
        </div>
    </div>
</div>
```
This follows the standard pattern library convention of wrapper/grid/row/box, doesn't require a brand new component with custom margins, and (in theory) just needs some colour swaps.

It turns out it's not that simple - [alt rows have a lot of custom and conditional styling](https://github.com/university-of-york/design-patterns/blob/master/src/sass/object/_grids.scss), some of which isn't appropriate for this component, so I think the current solution is probably better than the alternative. In general I still recommend leveraging the pattern library as much as possible before resorting to our own components.

public/stylesheets/styles.css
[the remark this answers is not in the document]
ce666-york: Resolved by the discussion below - there doesn't seem to be a suitable component we can copy and then just tweak the colours

ce666-york (approved):

FergusMcGlynn (approved):

RichardAllum (approved):

PR #107, opened 2021-02-10
Cfin 433 sticky footer

RichardAllum:


src/pages/index.js
```diff
@@ -19,35 +21,36 @@ import { HeroBanner } from "../components/HeroBanner";
 
 const App = ({ isSuccessfulSearch, searchResults, numberOfMatches, searchTerm }) => {
     return (
-        <>
-            <PageHead search={searchTerm} />
-            <UniversityHeaderWithSearch />
-            <UniversityTitleBar title="Courses" />
+        <UniversityWrapper>
+            <UniversityBody>
+                <PageHead search={searchTerm} />
```
FergusMcGlynn, line 26: Should `<PageHead>` sit inside `<UniversityBody>`? It outputs `<head>` tags, so I'd have thought it'd be more appropriate right at the top of the returned content.

src/pages/index.js
```diff
@@ -19,35 +21,36 @@ import { HeroBanner } from "../components/HeroBanner";
...
-            </WrappedMainGrid>
+                <HeroBanner>
+                    <Breadcrumbs screenBreadcrumbs={[{ label: "Course search", link: "/" }]} />
+                    <h1>Search undergraduate courses</h1>
+                    <Search
+                        searchTerm={searchTerm}
+                        numberOfMatches={numberOfMatches}
+                        numberOfResultsShown={searchResults?.length}
+                    />
+                </HeroBanner>
```
FergusMcGlynn, line 38: Minor thing - but git seems to think these lines for `<HeroBanner>` (and quite a few others in this file) have been deleted and recreated, whereas it seems from the content that the only change is indentation. Probably doesn't matter - either way, a `git blame` or viewing annotations would show that these lines were changed by this commit. I'm curious as to why it's doing this though - any idea?

src/pages/index.js
RichardAllum: Good point!

src/pages/index.js
RichardAllum: It's how github shows indentation changes I think, `git blame` isn't affected thankfully

ce666-york (approved):

src/pages/index.js
FergusMcGlynn: Yes, Github's default diff is not very helpful in this situation. I've discovered that you can add the URL param `?w=1` to the "Files changed" view URL and it will ignore whitespace in the diff. Doing this makes it look a lot easier to see the actual changes made.

FergusMcGlynn (approved):

ncn1 (approved):

PR #110, opened 2021-03-04
CFIN-445 Add sentence case function to capitalise Undergraduate tag

ChristianBeddows:


src/components/CourseDetails.js
```diff
@@ -2,12 +2,13 @@ import React from "react";
...
 import { RichTagList } from "./RichTagList";
 import { RichTag } from "./RichTag";
+import { sentenceCase } from "../utils/formatting";
 import PropTypes from "prop-types";
 
 const CourseDetails = ({ course }) => {
     return (
         <RichTagList>
-            <LevelAndAwardTag level={course.level} award={course.award} />
+            <LevelAndAwardTag level={sentenceCase(course.level)} award={course.award} />
```
ChristianBeddows, line 11: Do you think this is the right place to use this function?
I am unsure whether it belongs within the `LevelAndAwardTag` function, similar to how the startYear string manipulation happens within the `StartDateAndLengthTag` function on line 36.

mrgibbonsuoy: I'd agree with your position it keeps it consistent.

mrgibbonsuoy (approved):

src/components/CourseDetails.js
ChristianBeddows: I'm not sure if this is consistent - the function call should perhaps belong on line 23, similar to line 36, further down in this file

ncn1: Surely it's actually inconsistent - this one done in the _call to_ LevelAndAwardTag
 and startYear string manipulation happens _within_ the StartDateAndLengthTag.
 
 If we are saying the level description in a rich tag should always be in sentence case (which I think we are and was the implication of using css for it), I think it should also be used _within_ LevelAndAwardTag.

src/utils/formatting.js
```diff
@@ -0,0 +1,10 @@
+const sentenceCase = (string) => {
```
ncn1, line 1: Nice succinct and clear function!

src/components/CourseDetails.js
mrgibbonsuoy: I was thinking about someone opening the project and scanning the files. Whilst out of position you can see at a glance the tags.

ce666-york: Given it's a `LevelAndAwardTag` rather than a generic `Tag` component, I'd lean towards putting all functionality for rendering this tag correctly within the component. It's really not a problem though.

src/tests/utils/formatting.test.js
```diff
@@ -0,0 +1,7 @@
+import { sentenceCase } from "../../utils/formatting";
+
+describe("sentenceCase", () => {
+    it("Capitalises the first letter of the input string", async () => {
```
ce666-york, line 4: I'd like to see more tests for edge cases e.g. empty string, null string.

I think it would also be useful to put in tests to show what we expect to see when we receive input that's already in sentence case, input that's completely capitalised, and input with spaces/formatting. Pretend you don't know anything about the implementation details!

ce666-york (changes requested):
Small point about more tests!

src/tests/utils/formatting.test.js
FergusMcGlynn: This could probably be [shortened using Jest's `each` function](https://elfi-y.medium.com/reduce-boilerplate-test-code-with-jest-it-each-30a0eec9776d) - like the `where` clause in Spock tests. But it's not a problem how it is currently.

FergusMcGlynn (approved):

src/components/CourseDetails.js
FergusMcGlynn: I'm a little confused - is there any reason not to call `sentenceCase()` in `RichTag`, so that it applies to all text in all tags no matter which type? E.g, in `RichTag.js`:
```
    <li className="rich-tag">                                                                                     
        {topIcon && <i aria-hidden className={`c-icon c-icon--${topIcon} c-icon--2x`} data-testid="tag-icon" />}  
        <p className="rich-tag__title">{sentenceCase(title)}</p>                                                                
        <p className="rich-tag__text">{sentenceCase(subText)}</p>                                                               
    </li>                                                                                                         
```

ce666-york (approved):

src/components/CourseDetails.js
ChristianBeddows: I like the idea of using the function within `RichTag` so that it always applies to text - though following from our discussion earlier I'm going to leave it as it is, in the idea of changing as little as possible within this changeset. I think it could be something to revisit though!

PR #114, opened 2021-04-19
CFIN-450 Make the accessibility tree course links be the course title and award

ncn1:
The course links until this change have appeared in the accessibility tree with link text containing the course's title and all of its tags eg. 'Music - BA (Hons) Undergraduate BA (Hons) Start 2021 3 years full-time'

This change makes it 'Music - BA (Hons)'.

With grateful acknowledgement of Fergus's detailed notes on the JIRA story of how to tackle this.

### DIFF
--- a/src/components/Course.js
+++ b/src/components/Course.js
@@ -11,12 +11,12 @@ const Course = ({ course }) => {
         return course.title;
     };
 
-    const linkId = "course-title-" + title(course).replace(/ /g, "-");
+    const headingId = "course-title-" + title(course).replace(/ /g, "-");
 
     return (
         <li className="c-listings-item ">
-            <a className="c-listings-item__link" aria-labelledby={linkId} href={course.liveUrl}>
-                <h2 id={linkId} className="c-listings-item__title">
+            <a className="c-listings-item__link" aria-labelledby={headingId} href={course.liveUrl}>
+                <h2 id={headingId} className="c-listings-item__title">
                     {title(course)}
                 </h2>
                 <div className="c-listings-item__description">
--- a/src/components/Course.js
+++ b/src/components/Course.js
@@ -11,10 +11,14 @@ const Course = ({ course }) => {
         return course.title;
     };
 
+    const linkId = "course-title-" + title(course).replace(/ /g, "-");
+
     return (
         <li className="c-listings-item ">
-            <a className="c-listings-item__link" href={course.liveUrl}>
-                <h2 className="c-listings-item__title">{title(course)}</h2>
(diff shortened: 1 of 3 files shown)


src/tests/components/Course.test.js
```diff
@@ -42,4 +42,29 @@ describe("Course", () => {
...
+        const course = {
+            title: "A Course",
+            liveUrl: "https://fakecourse.notadomain/",
+            level: "undergraduate",
+            award: "Award",
+        };
+
+        render(<Course course={course} />);
+
+        expect(screen.getByRole("link", { name: "A Course - Award" })).toBeInTheDocument();
```
RichardAllum, line 56: The `getByRole` here is a touch confusing for me, does the `a` tag have an implicit `link` role. or is it granted to the `h2` by the `aria-labelledby` attribute?

ncn1: The a tag has an implicit link role.  Then the aria-labelledby is just telling it to use the heading bit for the aria label.

src/components/Course.js
```diff
@@ -11,10 +11,14 @@ const Course = ({ course }) => {
         return course.title;
     };
 
+    const linkId = "course-title-" + title(course).replace(/ /g, "-");
```
FergusMcGlynn, line 14: I don't think course titles are unique - `Mathematics - BSc (Hons)` and `Mathematics - MMath (Hons)` have the same course title, `Mathematics`. This line of code would give them both the same `id`, which might things up for assistive tools. I think we should try to make the `id` unique for each course entry.

Also, v minor thing, the variable is called `linkId`, but it's actually used as an id for the `<h2>` heading rather than the link, so perhaps it should be `headingId`?

FergusMcGlynn (changes requested):
I think we should try to make the new `id` unique for each course. Would combining title and award make a unique combination? Might be worth asking BAs or Digital Platforms perhaps?

src/components/Course.js
ncn1: title(course) is referencing the function above, which includes the award in the title if there is one present - drawback of only being shown the changed code. Happy to change the id name.

ncn1:
> I think we should try to make the new `id` unique for each course. Would combining title and award make a unique combination? Might be worth asking BAs or Digital Platforms perhaps?

It's already doing this - see response to comment.

FergusMcGlynn (approved):
Thanks Nicola - I didn't realise that the award was included in the return value from `title()`. Looks good!

RichardAllum (approved):

PR #117, opened 2021-04-23
CFIN-451 Add most basic 'main' element

RichardAllum:
We aren't using the HTML <main> due to a lack of global support.

### DIFF
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -29,36 +29,38 @@ const App = ({ isSuccessfulSearch, searchResults, numberOfMatches, searchTerm })
                     <UniversityHeaderWithSearch />
                     <UniversityTitleBar title="Courses" />
 
-                    <HeroBanner>
-                        <Breadcrumbs screenBreadcrumbs={[{ label: "Course search", link: "/" }]} />
-                        <h1>Search undergraduate courses</h1>
-                        <Search
-                            searchTerm={searchTerm}
-                            numberOfMatches={numberOfMatches}
-                            numberOfResultsShown={searchResults?.length}
-                        />
-                    </HeroBanner>
+                    <div role="main">
+                        <HeroBanner>
+                            <Breadcrumbs screenBreadcrumbs={[{ label: "Course search", link: "/" }]} />
+                            <h1>Search undergraduate courses</h1>
+                            <Search
+                                searchTerm={searchTerm}
+                                numberOfMatches={numberOfMatches}
+                                numberOfResultsShown={searchResults?.length}
+                            />
+                        </HeroBanner>
 
-                    <WrappedMainGrid>
-                        <GridRow>
-                            <GridBoxFull>
-                                <SearchResultsDescription
-                                    searchTerm={searchTerm}
-                                    numberOfMatches={numberOfMatches}
-                                    numberOfResultsShown={searchResults?.length}
-                                />
-                            </GridBoxFull>
-                        </GridRow>
-                        <GridRow>
-                            <GridBoxFull>
-                                <CourseSearchResults
-                                    isSuccessfulSearch={isSuccessfulSearch}
-                                    searchResults={searchResults}
-                                    searchTerm={searchTerm}
-                                />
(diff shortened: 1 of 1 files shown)


RichardAllum:
We will definitely want to test this on as many browsers as possible to ensure that the `main` div doesn't have any styles applied or mess up the page in any way.

RichardAllum:
Before pulling 'dev' into this branch the percy tests failed showing no footer on firefox or chrome for all widths, I'm unable to replicate this locally on firefox or chrome so I think this may have been a glitch given the current pass. It's still worth doing manual cross-browser testing once this hits staging if it passes review.

ce666-york (approved):

src/pages/index.js
```diff
@@ -29,36 +29,38 @@ const App = ({ isSuccessfulSearch, searchResults, numberOfMatches, searchTerm })
...
-                    </HeroBanner>
+                    <div role="main">
+                        <HeroBanner>
+                            <Breadcrumbs screenBreadcrumbs={[{ label: "Course search", link: "/" }]} />
+                            <h1>Search undergraduate courses</h1>
+                            <Search
+                                searchTerm={searchTerm}
+                                numberOfMatches={numberOfMatches}
+                                numberOfResultsShown={searchResults?.length}
+                            />
```
ncn1, line 40: I ran this and noticed there were several 'generic' nodes before the actual search section was reached. I think the role should be lower down, to cut out the breadcrumbs (which should probably have a navigation role - another story for a pattern library change required for that).
```suggestion
                        <div role="main">
                            <h1>Search undergraduate courses</h1>
                            <Search
                                searchTerm={searchTerm}
                                numberOfMatches={numberOfMatches}
                                numberOfResultsShown={searchResults?.length}
                            />
                        </div>

```
You will also need to remove the existing div="main" - couldn't work ouot how to include all the required code change.

ce666-york: Would this suggestion move the search results out of the main section?

It may be there's no perfect solution here, but I think the main section should include both the course search form and the search results - to me this is the "primary content of the document".

RichardAllum: Yeah, it's unfortunate that we need to contain the breadcrumbs in order to contain the search results - it might be possible to put them closer together in the DOM if we were to do some horrible things to the CSS of the hero banner, but this is the only way right now.

ncn1: Is that how it works? It has to contain the results? I was assuming it was more like something the tool can jump to the start of, but doesn't care how much stuff is contained within it. A bit like a 'skip to content' link.

ce666-york: [This page](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Main_role) indicates that it contains all the main content, which would make sense semantically. It also mentions "skip to content", but only as a way to skip to the `main` section 🤷‍♀️

FergusMcGlynn: Would it be an option to take the `Breadcrumb` out of the `HeroBanner` (i.e. move it above), but style it with the same styling as the `HeroBanner` so that to a regular browser it looks like there's been no change? Then we'd have separated the navigation stuff from the main stuff, which seems a little cleaner to me, and maybe to assistive tools as well?

ChristianBeddows (approved):

src/pages/index.js
RichardAllum: This is possible, but I'm not sure about the side effects. It requires adding another o-wrapper above the hero banner with a hardcoded style, so it would not change if the hero banner styles change (the hero banner uses a slight gradient, so just repeating the styles results in an ugly colour shift). I think if we would need to do this then we may need to reconsider the hero banner, which might become necessary if we need the side menus anyway.

I recommend that we take the version from this commit into the user testing, and if we get feedback that the breadcrumbs are in the way then we should separate them then. We may also get feedback about side menus, allowing us to make the decisions related to the hero banner together.

FergusMcGlynn (approved):

PR #120, opened 2021-04-27
CFIN-438 Add Undergraduate Courses page styling

ce666-york:
First of all I apologise for the number of changes in this PR - I've run out of time! Thankfully they're all straightforward, I think.

- Added menu navigation
- Added breadcrumbs
- Added side menu ("subnavigation")
- Added coronavirus notice
- Updated page heading

For the breadcrumbs and side menu I'd like to go back and put these in the React pattern library at some point, particularly the changes to breadcrumbs (which involve allowing users to pass through arbitrary properties, and making it a `nav` instead of a `div`).

For now this review places the side menu below the hero banner, and encapsulates all of the search bar, search results and side menu in the `main` content (option 2-1 in the Slack conversation). This may change depending on the feedback 📝


ChristianBeddows (approved):
I think the sidebar looks good but I find it a little jarring that the search bar is a different width to the results section

ce666-york:
I've removed the side menu based on feedback, check out the Percy snapshot 🐷

src/components/UndergraduateMenuNavigation.js
```diff
@@ -0,0 +1,59 @@
+import { MenuNavigation } from "@university-of-york/esg-lib-pattern-library-react-components";
```
ncn1, line 1: When I checkout the CFIN-438 branch, and run in dev I see the following 
image
I then ran the menu navigation test, and it failed, as only the following was rendering
```
<DocumentFragment>
  <nav
    class="c-nav c-nav--main"
    role="navigation"
  />
</DocumentFragment>
```

ncn1: Doh! Ignore, needed to run npm install!

ncn1 (approved):
From a UX point of view, this makes it a very busy heading, with the coronavirus notice and the feedback banner and takes up all of my 15 inch laptop screen, so have to scroll down to see results - be better once we don't need the Coronavirus notice. Mobile view OK, though, because you expect to scroll down. And hero banner especially good as does highlight where the real functionality starts.

PR #121, opened 2021-05-04
Use calendar icon for course start rather than clock

FergusMcGlynn:
This came up at the stakeholder meeting - Paul Armstrong requested this change, and I tend to agree that a calendar is more appropriate for a date than a clock. 

Here's how it looks:
image

What do we think?


itspaulkelly:
Yes I agree - we've used the calendar icon in other places where we're indicating a date, eg the term dates link from https://www.york.ac.uk/about/

ce666-york (approved):

PR #126, opened 2021-05-07
CFIN-423 Add cookie banner

ncn1:


src/tests/pages/index.test.js
```diff
@@ -58,6 +59,23 @@ describe("App", () => {
...
+        render(<App />);
+
+        expect(screen.getByText("our use of cookies")).toBeVisible();
+    });
+
+    it("hides cookie banner when dismiss clicked", async () => {
+        cleanCookies();
+        render(<App />);
+        const icons = document.querySelectorAll("i");
+        fireEvent.click(icons[3]);
```
RichardAllum, line 74: Looking good, would it be better to find the icon inside the banner in case the number of icons on the screen changes though?

RichardAllum (approved):

PR #130, opened 2021-05-12
Update dependencies (including lighthouse major version)

ce666-york:
A quick PR to update the lighthouse major version, which should resolve the outstanding security alert. I also ran an `npm update` (and `npm update --dev`) to update all minor/patch versions of dependencies. Application runs locally, all tests pass and performance checks run.


ce666-york:
Want to look into why the package size has increased before merging

ncn1 (approved):

ce666-york:
I managed to trim down the increased package size by ~1.5MB by removing a couple of dependencies and some unnecessary repo files included in many dependencies (READMEs, tests etc.), but it's still 2.25MB more than before. I think this is simply due to the nature of the updated dependencies, so I'm going to leave things here.

ChristianBeddows (approved):

FergusMcGlynn:
I'm also a little perplexed by the change in package size as a result of updating Lighthouse. The bot that reports the size change seems to be a little off to me - the most recent PR before this one reported a package size of 12.3MB, and for this one it states the size has increased by 2.25MB to a new total of 17MB - those numbers don't seem quite right. Lighthouse, being a dev dependency, shouldn't be affecting the size of the package that Serverless builds, so I'm basically confused as to what's going on here.

ce666-york:
@FergusMcGlynn I did also run `npm update` which updates all packages to their latest minor/patch versions!

RichardAllum (approved):

FergusMcGlynn (approved):
@ce666-york Ah, OK, that could well be it. I'm still confused about 12.3MB + 2.25MB = 17MB - but maybe that's not worth worrying about too much. I'm a little concerned that the general bloating of dependencies over time means that this app is going to get bigger and slower. I guess if we see the performance scores drop then we'll need to come up with a workaround in due course. We're still quite a way off the 50MB limit that AWS imposes on package sizes for Lambdas, which is the main thing we need to avoid.

PR #132, opened 2021-05-19
Condense the README and add GitHub status badges

ce666-york:
Our README is very detailed, but I've felt overwhelmed by it for a while - times have moved on and we have more knowledge of AWS and npm, and the focus should be on being a getting started guide rather than a go-to for every single detail.

This PR is a (perhaps bold) suggestion to trim down the README to some of the essentials, with some information being moved to the GitHub wiki, and some being removed altogether.

At the same time I decided to tackle the GitHub workflows, opting to split out the checks into separate workflows - this gives us more visibility of exactly what goes wrong in a build, at the trade-off of spending more GitHub build minutes. I think it's worth the cost.

Splitting out the checks also makes it very easy to have a clean set of GitHub status badges that immediately tell us the health of the repo. These show the result of the checks being run against the `dev` branch.

You can see the rendered version of the readme [here](https://github.com/university-of-york/uoy-app-course-search/tree/update-readme). You can see how the updated checks appear by looking at this PR.

Thoughts welcome :slightly_smiling_face:

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ npm test
 
 ### Visual tests
 
-We use [Percy](https://percy.io/ad91c322/uoy-app-course-search) for visual testing. This adds snapshots of UI changes to each pull request. More detail can be found in the York Wiki Service page (University users only): [Testing: Percy, automation in testing with GitHub](https://wiki.york.ac.uk/pages/viewpage.action?pageId=220921899)
+We use [Percy](https://percy.io/ad91c322/uoy-app-course-search) for visual testing. This adds snapshots of UI changes to each pull request.
 
 ### Performance tests
 
--- a/README.md
+++ b/README.md
@@ -52,7 +52,12 @@ Deployment to the development and production environments happens through GitHub
 
 ### Deploying to your own AWS account
 
-You can run Course Search in your own AWS account. Make sure you've got an active token under `~/.aws/credentials` (e.g. by logging into your account with `saml2aws`) and then run:
+You can run Course Search in your own AWS account. You will need to do the following:
+
+- Make sure you've got an active token under `~/.aws/credentials` (e.g. by logging into your account with `saml2aws`)
+- Define an environment variable called `AWS_ACCOUNT_ID` populated with your AWS account ID
+   
+Then run:
 
 ```
 npm run deploy:dev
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -1,278 +1,99 @@
...
-##### Intellij
-
-You can configure Intellij to format code, using `prettier`, when you save a file and when 
-you run Intellij's formatting command (`Ctrl-Alt-L`). To do this:
-* install the `prettier` plugin (under `File` > `Settings` > `Plugins`)
-* go to `File` > `Settings` > `Languages & Frameworks` > `Javascript` > `Prettier` and
-check the options `on save` and `on code reformat`
-
-To make Intellij use the `prettier` formatting rules while you edit code, open
-`package.json` and above the code window it will prompt you to `Use code style based on prettier for this project?`
```
ncn1, line 176: This is useful info for me still, especially the IntelliJ stuff which I have done once and forgotten how to do should I need to set it up again. A link to somewhere else with this info would therefore be helpful.

The formatting stuff is also useful in the light of my experience using the React Pattern Library, where I could only check the code formatting once it was staged for commit - not something you can do directly in IntelliJ.

README.md
```diff
@@ -1,278 +1,99 @@
...
-##### Developer AWS sandbox
-
-When running Course Search in your AWS sandbox, use `npm run deploy` to deploy using
-production environment variables (and therefore the production Courses API), and 
-`npm run deploy:dev` to deploy using development environment variables (and
- therefore the development Courses API).
+### Visual tests
 
-##### courses.dev.app.york.ac.uk
+We use [Percy](https://percy.io/ad91c322/uoy-app-course-search) for visual testing. This adds snapshots of UI changes to each pull request.
```
ncn1, line 49:
```suggestion
We use [Percy](https://percy.io/ad91c322/uoy-app-course-search) for visual testing. This adds snapshots of UI changes to each pull request. More detail can be found in the York Wiki Service page (University users only): [Testing: Percy, automation in testing with GitHub](https://wiki.york.ac.uk/pages/viewpage.action?pageId=220921899)
```

ncn1: Sorry, I seem to have committed this - I misunderstood 'Commit suggestion' as 'Submit suggestion'. Hope it's OK to add back in - I would waste time searching for this information without it.

ChristianBeddows (approved):
Love the badges! <3

README.md
ce666-york: Good feedback, I think a wiki page is more suitable for IntelliJ-specific instructions, I'll move it over to one

README.md
ce666-york: I'm conscious that this is a public repo, I don't think it's a great experience to have private information linked in a public README which is why I initially took it out. I still feel that way, is there another solution here?

FergusMcGlynn (approved):

README.md
ncn1: Perhaps a page under 'Developer Guides' in this repo wiki?

README.md
ncn1: Unlike the Percy stuff, the prettier IntelliJ stuff would sit better on the York wiki, and no need to link from here. The formatting stuff about running separately to the tests is specific to this repo, so could stay here or go in the repo wiki.

PR #134, opened 2021-06-04
Retain the level=undergraduate parameter when searching

RichardAllum:
This PR also adds [cypress](https://docs.cypress.io/) to the application. The workflow takes a long time, we might not need to install everything beforehand if it's done in the cypress action.


RichardAllum:
The linting fails as many functions aren't declared in the submit_form functional test, renaming to `submit_form.test.js` fixes some of the issues but it still wants `context` and `cy` to be defined - importing these messes with intellij's linting, however. Would it be better to exclude the cypress folder from xo?

src/tests/components/Search.test.js
```diff
@@ -22,4 +22,11 @@ describe("Search", () => {
 
         expect(getByRole("textbox", { name: "Search" }).value).toEqual("French");
     });
+
+    //
```
ce666-york, line 26:
```suggestion
```

.github/workflows/functional-tests.yml
```diff
@@ -0,0 +1,32 @@
+name: Functional Tests
+
+on:
+  workflow_dispatch:
+  pull_request:
```
ce666-york, line 5: How about we run these on merge too, to match the other workflows?

```suggestion
  pull_request:
  push:
    branches:
      - dev
```

ce666-york (commented):
Looking at our [XO config](https://github.com/university-of-york/uoy-app-course-search/blob/dev/.xo-config.json), we could investigate whether we can add an override for functional tests with a `cypress` env or similar that allows the global `cy` and `context` - my understanding is that's what the `jest` override does with `describe` and `it` etc.

ce666-york (approved):

cypress/integration/submit_form.js
```diff
@@ -0,0 +1,16 @@
...
+
+    describe("visiting the search", () => {
+        it("submits a level parameter when searching", () => {
+            cy.get("input[name=search]").type("maths").should("have.value", "maths");
+
+            cy.get("#course-search-form").submit();
+            cy.url().should("include", "level=undergraduate");
+        });
+    });
+});
```
FergusMcGlynn, line 16: Is there any reason not to put this file/folder somewhere underneath `/src/tests/`? I think it'd be good to keep the bits of code associated with testing in roughly the same place if possible.

src/tests/components/Search.test.js
```diff
@@ -22,4 +22,10 @@ describe("Search", () => {
 
         expect(getByRole("textbox", { name: "Search" }).value).toEqual("French");
     });
+
+    it("contains a level parameter for comparing analytics", () => {
```
FergusMcGlynn, line 26: I didn't quite understand this description on first reading. Maybe we could make it slightly clearer?

```suggestion
    it("contains a 'level' query parameter for comparing analytics between this app and Course Finder", () => {
```

cypress/integration/submit_form.js
RichardAllum: Nothing except this is the happy path for Cypress, I'd be happy to add a tech debt task to the backlog to try and move it in future though

FergusMcGlynn: Thanks Richard - though not sure I quite understand what you mean by "happy path" here. Do you mean that we would need to add additional config to the app in order for this test file to sit under `/src/tests`?

FergusMcGlynn: It looks like [some additional config would be needed](https://stackoverflow.com/a/49196174), although not heaps. The [Cypress documentation](https://docs.cypress.io/guides/core-concepts/writing-and-organizing-tests#Folder-structure) states 
```
While Cypress allows you to configure where your tests, fixtures, and support files are located, if you're starting your first project, we recommend you use the [default] structure.
```
If no one else thinks this test file should live under `/src/tests/` then I'm happy for it to stay as it is.

ncn1: I would prefer it with the other tests as well as I can see myself wasting time trying to find it in future.

RichardAllum: That's fine, we can move it. In order to get xo to work we'll need to have a different file naming system from our other tests, would it be better to have `tests/{cypress|functional}/submit-form.js` or `tests/something/submit_form.cypress.js`?

FergusMcGlynn: I'd be fine with `src/tests/cypress/submit-form.js`. I think since the tests are written using in Cypress-specific syntax, it'd fine to name the folder that contains the test `cypress`.

FergusMcGlynn (approved):

PR #136, opened 2021-06-09
CFIN-520 Log all requests to course search

RichardAllum:
The logic to get the client and Source IPs is different to that in the courses API, more closely matching the AuditLogger that Ian created.

### DIFF
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -100,7 +100,7 @@ const getServerSideProps = async (context) => {
     const { isSuccessfulSearch, searchResponseData } = await searchForCourses(searchTerm);
 
     if (!isSuccessfulSearch) {
-        console.log(logEntry(context.req, LOG_TYPES.ERROR, context.query));
+        console.log(logEntry(context.req, LOG_TYPES.ERROR, context.query, { results: searchResponseData.results }));
     }
 
     return {
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -87,7 +87,7 @@ App.propTypes = {
 const getServerSideProps = async (context) => {
     const searchTerm = context.query.search === undefined ? context.query.q : context.query.search;
 
-    console.log(logEntry(context.req, LOG_TYPES.AUDIT));
+    console.log(logEntry(context.req, LOG_TYPES.AUDIT, context.query));
 
     if (noSearchConducted(searchTerm)) {
         return { props: {} };
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -23,6 +23,8 @@ import { UndergraduateMenuNavigation } from "../components/UndergraduateMenuNavi
 import { UndergraduateBreadcrumbs } from "../components/UndergraduateBreadcrumbs";
 import { CoronavirusNotice } from "../components/CoronavirusNotice";
 import { GlobalNotice } from "../components/GlobalNotice";
+import { logEntry } from "../utils/logEntry";
+import { LOG_TYPES } from "../constants/LogTypes";
 
 const App = ({ isSuccessfulSearch, searchResults, numberOfMatches, searchTerm }) => {
     return (
@@ -84,6 +86,8 @@ App.propTypes = {
 
 const getServerSideProps = async (context) => {
     const searchTerm = context.query.search === undefined ? context.query.q : context.query.search;
+
+    console.log(logEntry(context.req, LOG_TYPES.AUDIT));
 
     if (noSearchConducted(searchTerm)) {
         return { props: {} };
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -87,7 +87,7 @@ App.propTypes = {
 const getServerSideProps = async (context) => {
     const searchTerm = context.query.search === undefined ? context.query.q : context.query.search;
 
-    console.log(logEntry(context.req, LOG_TYPES.AUDIT, context.query));
+    console.info(logEntry(context.req, LOG_TYPES.AUDIT, context.query));
 
     if (noSearchConducted(searchTerm)) {
         return { props: {} };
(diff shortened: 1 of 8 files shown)


src/utils/logEntry.js
```diff
@@ -0,0 +1,49 @@
+/**
+ *
+ * @param {http.IncomingMessage} request
+ * @param {String} logType
+ * @param {Object} additionalDetails Added directly to the log
+ */
```
FergusMcGlynn, line 6: Not a problem - just noticing the JSDoc for this function, which we don't have anywhere else in this project (as far as I'm aware). I know that Ian is a fan of JSDoc, but that Natacha is not a fan. I don't really have an opinion, just wondered if it's something we want to decide to either adopt as a team or not adopt? I guess I'd be more comfortable if we decided to us it consistently from now on, or not use it, rather than potentially using it ad hoc.

RichardAllum: I am personally a fan but I know as a team we don't tend to use them, in this case I thought it was important to note the type of the `request` variable as it is something that may not be obvious at a glance. If this was to be turned into a library I would expect functions that convert various types of request/object into a common format that can be logged, and that would make this potentially useful.

src/tests/utils/logEntry.test.js
```diff
@@ -0,0 +1,94 @@
...
+                    type: "GET",
+                    statusCode: StatusCodes.OK,
+                    version: "v1",
+                },
+                sensitive: false,
+                schemaURI: "https://github.com/university-of-york/uoy-app-course-search",
+                type: "audit",
+                queryStringParameters: {
+                    search: "biology",
+                },
```
ce666-york, line 74: These don't appear in the log message when running locally; do they when running in AWS?

```
{"timestamp":"2021-06-10T10:12:17.689Z","ip":{"client":null,"source":null,"sourcePort":null},"req":{"user":null,"service":"uoy-app-course-search"},"correlationId":null,"self":{"application":"uoy-app-course-search","type":"GET","stat
usCode":null,"version":"v1"},"sensitive":false,"schemaURI":"https://github.com/university-of-york/uoy-app-course-search","type":"audit"}
```

src/pages/index.js
```diff
@@ -85,6 +87,8 @@ App.propTypes = {
 const getServerSideProps = async (context) => {
     const searchTerm = context.query.search === undefined ? context.query.q : context.query.search;
 
+    console.log(logEntry(context.req, LOG_TYPES.AUDIT));
```
ce666-york, line 90: Suggest we also log an error further down if `isSuccessfulSearch` is `false`

ce666-york: I think it's worth it, it handles any situation where the Courses API error logging isn't reached, and it gives more context in the Course Search logs

src/tests/utils/logEntry.test.js
RichardAllum: Good find, I've fixed that now

src/utils/logEntry.js
```diff
@@ -0,0 +1,51 @@
+/**
+ *
+ * @param {http.IncomingMessage} request
+ * @param {String} logType
+ * @param {Object} queryParameters
+ * @param {Object} additionalDetails Added directly to the log
+ */
+
+const logEntry = (request, logType, queryParameters, additionalDetails) => {
```
ncn1, line 9: A test to prove additionalDetails are added to the log would be good, now you are using it for error reporting.

src/pages/index.js
```diff
@@ -95,6 +99,10 @@ const getServerSideProps = async (context) => {
 
     const { isSuccessfulSearch, searchResponseData } = await searchForCourses(searchTerm);
 
+    if (!isSuccessfulSearch) {
+        console.log(logEntry(context.req, LOG_TYPES.ERROR, context.query, { results: searchResponseData.results }));
```
ce666-york, line 103: :slightly_smiling_face: 

```suggestion
        console.error(logEntry(context.req, LOG_TYPES.ERROR, context.query, { results: searchResponseData.results }));
```

ce666-york (approved):

src/pages/index.js
```diff
@@ -95,6 +99,10 @@ const getServerSideProps = async (context) => {
 
     const { isSuccessfulSearch, searchResponseData } = await searchForCourses(searchTerm);
 
+    if (!isSuccessfulSearch) {
+        console.error(logEntry(context.req, LOG_TYPES.ERROR, context.query, { results: searchResponseData.results }));
+    }
```
FergusMcGlynn, line 104: Apologies for not mentioning this before... Given that we've modified the code in `index.js`, I think it'd be good to add a test (or two, given that there are 2 changes) to `index.test.js` to verify that the code does what you intend. Something along the lines of
```
    it("logs an error when the course search failed", async () => {
        console.error = jest.fn();
        
        searchForCourses.mockResolvedValue({
            isSuccessfulSearch: false,
            searchResponseData: { numberOfMatches: 0, results: [] },
        });

        const response = await getServerSideProps(contextWithSearchTerm);

        expect(console.error).toHaveBeenCalledTimes(1);
        expect(console.error).toHaveBeenCalledWith(.....);
    });
```

ChristianBeddows (approved):

src/pages/index.js
```diff
@@ -85,6 +87,8 @@ App.propTypes = {
 const getServerSideProps = async (context) => {
     const searchTerm = context.query.search === undefined ? context.query.q : context.query.search;
 
+    console.log(logEntry(context.req, LOG_TYPES.AUDIT, context.query));
```
FergusMcGlynn, line 90: Minor thing, but maybe consider `console.info()` rather than `console.log()`? They're functionally equivalent, as far as I can tell, but personally I find using the `info`/`warn`/`error` logging levels help when reading code to understand what's going on (whereas `console.log()` is often used as a catch-all).

src/pages/index.js
FergusMcGlynn: I've created [a branch containing tests that I think it'd be good to include](https://github.com/university-of-york/uoy-app-course-search/compare/CFIN-520...CFIN-520_add_index_tests). No obligation to use them, of course!

FergusMcGlynn (approved):

PR #138, opened 2021-06-10
CFIN-524: Prepare Course Search for hosting on www.york.ac.uk

FergusMcGlynn:
We're planning to host Course Search at https://www.york.ac.uk/courses. The application will still retain its current domain name - courses.app.york.ac.uk - and will still be accessible from there, but we will advertise https://www.york.ac.uk/courses to users.

We also want to take advantage of CloudFront caching that is enabled for all off-campus visitors to www.york.ac.uk. This will mean that users generally receive cached versions of the main page and static assets from their nearest AWS edge location rather than from the application itself (the API + Lambda in Dublin).

To do these, a couple of minor changes are needed to Course Search:

* move the application's base path from `/` to `/courses`
   * this means that the application URL will be https://courses.app.york.ac.uk/courses
   * I have added in a redirect from `/` to `/courses` so that if you go to https://courses.app.york.ac.uk you'll be automatically redirected to https://courses.app.york.ac.uk/courses
      * This redirect is configured in ~~`server.js`~~ [next.config.js](https://nextjs.org/docs/api-reference/next.config.js/redirects) - I finally figured out how to make it work
      * This redirect ~~therefore only works in AWS~~ works both locally (i.e. http://localhost:3000 will automatically redirect to http://localhost:3000/courses) and in AWS
* enable caching of pages and static assets
   * I have set these to be cached for 1 day


ncn1 (approved):

README.md
```diff
@@ -62,6 +62,16 @@ We use the [eslint-plugin-jsx-a11y](https://github.com/jsx-eslint/eslint-plugin-
 
 To reinforce an accessibility-first approach, we use ARIA roles and reference these in automated tests where possible.
 
+### Functional tests
+
+We use [Cypress](https://www.cypress.io) for functional tests. To execute them, start the application locally and then run
+
+```
+npx cypress open
```
ce666-york, line 70: Word of warning that this doesn't work when using the Windows Subsystem for Linux, it just hangs (same for running headless with `npx cypress run`). I decided to install Node in my Windows environment and run it from there, which worked fine after doing `npx cypress install`

ce666-york (approved):

README.md
FergusMcGlynn: Thanks Chris - I have decided to take the Cypress bit of the README out of this pull request and will create a new PR for it. It needs a little more thought, and I wasn't completely comfortable with including it in this PR anyway.

FergusMcGlynn:
I have modified the code in this PR - the redirect from "/" to "/courses" is now done by Next within the Course Search app, rather than by `server.js` outside the app. The effect is largely the same, but it means that the redirect works when running the app locally, which wasn't the case when it was done in `server.js`. Therefore fewer changes are needed to make the new `/courses` base path work.

ce666-york (approved):

ncn1 (approved):
The next redirect does look much simpler - good work.

ChristianBeddows (approved):

PR #141, opened 2021-06-30
Replace performance check script with action

ChristianBeddows:
As my recent CPD project I pulled the lighthouse performance check into it's own [dedicated action](https://github.com/university-of-york/esg-action-performance-check), allowing for it to be used within any project - requiring only a handful of parameters in the workflow file instead of a whole test script within the repo.

It also aims to make the results of the check more visible - see the conversation for this PR! 
The reports are uploaded as an artifact by the workflow - you can view the reports by downloading this from the Check's details


ncn1 (approved):
I think it would be good to do a short session on the new action at the ESG developers meeting to publicise to the other teams.

FergusMcGlynn (approved):
Great work! 🎉

ce666-york (approved):
Send a PR round for the action itself too! 👀

PR #151, opened 2021-09-09
CFIN-550 Update breadcrumbs to 2022/23

ce666-york:
Just a small change to update the breadcrumbs to 2022/23 to align with the existing Course Finder

### DIFF
--- a/src/components/UndergraduateBreadcrumbs.js
+++ b/src/components/UndergraduateBreadcrumbs.js
@@ -11,7 +11,7 @@ const UndergraduateBreadcrumbs = () => {
                 <span className="c-breadcrumb__separator">&gt;</span>
                 <a href="https://www.york.ac.uk/study/undergraduate/">Undergraduate</a>
                 <span className="c-breadcrumb__separator">&gt;</span>
-                <a href="https://www.york.ac.uk/study/undergraduate/courses-2021/">Courses 2021/22</a>
+                <a href="https://www.york.ac.uk/study/undergraduate/courses/">Courses 2022/23</a>
             </div>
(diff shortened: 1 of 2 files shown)


src/components/UndergraduateBreadcrumbs.js
```diff
@@ -11,7 +11,7 @@ const UndergraduateBreadcrumbs = () => {
                 <span className="c-breadcrumb__separator">&gt;</span>
                 <a href="https://www.york.ac.uk/study/undergraduate/">Undergraduate</a>
                 <span className="c-breadcrumb__separator">&gt;</span>
-                <a href="https://www.york.ac.uk/study/undergraduate/courses-2021/">Courses 2021/22</a>
+                <a href="https://www.york.ac.uk/study/undergraduate/courses/">Courses 2022/23</a>
```
FergusMcGlynn, line 14: Given that we only ever present 1 years worth of courses at a time, and that the link no longer has a year embedded in it, and that each course displays the year that it starts, would it be worth just changing this breadcrumb string to `Courses`?

ChristianBeddows (approved):

src/components/UndergraduateBreadcrumbs.js
ce666-york: In the short term for A/B testing showing 2022/23 is what we'd like to do to be consistent with Course Finder. I think in the longer term you're right, also because in future we'll be displaying multiple academic years (some work we'll need to do to properly support confirmation/clearing next year).

FergusMcGlynn (approved):

PR #160, opened 2021-11-23
Add instructions on accessing app after sandbox deployment

FergusMcGlynn:
When deploying Course Search to a sandbox account, it's not obvious what URL it can be accessed from, nor why the styling doesn't look right. This pull requests adds some text to the README to clarify these issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,25 @@ npm run deploy:dev
 npm run deploy
 ```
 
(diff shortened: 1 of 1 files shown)


rmil (approved):
Looks good to me :)

RichardAllum (approved):
Looks good!

PR #161, opened 2021-11-23
CFIN-582: Logger

rmil:
Pretty barebones PR, so hopefully won't be too bad, as a little summary:
- Added Pino
- Added request-ip
- Log format

## Added Pino

This has introduced a new `logger` object replacing all `console.*`.

This has the perk of taking inputs as JSON so we can remove `JSON.stringify` and make our tests more type safe as we can replace `expect.stringContaining` with `expect.objectContaining`.

When an error occurs, previously it would be popped inside the `additionalDetails` field inside `logEntry`. Error now has it's own in `logEntry` and since Pino prefers a specific field to be error and it matches the log format.

## Added request-ip
This library parses a given HTTP request and finds the client's IP through multiple headers.

## Log format
Now logging in the team's new decided [format](https://wiki.york.ac.uk/pages/viewpage.action?spaceKey=ittechdocs&title=AWS+Application+logging+fields) rather than the previous audit log policy.

## Checklist
- [x] Tests passing
- [x] Coverage :100:
- [x] Acceptance criteria looks like it's met
- [x] Works locally
- [x] Appears to work in limited sandbox testing

### DIFF
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -22,8 +22,8 @@ import { SearchResultsDescription } from "../components/SearchResultsDescription
 import { UndergraduateMenuNavigation } from "../components/UndergraduateMenuNavigation";
 import { UndergraduateBreadcrumbs } from "../components/UndergraduateBreadcrumbs";
 import { GlobalNotice } from "../components/GlobalNotice";
+import { logger } from "../utils/logger";
 import { logEntry } from "../utils/logEntry";
-import { LOG_TYPES } from "../constants/LogTypes";
 
 const App = ({ isSuccessfulSearch, searchResults, numberOfMatches, searchTerm }) => {
     return (
@@ -85,11 +85,7 @@ App.propTypes = {
 const getServerSideProps = async (context) => {
     const searchTerm = context.query.search === undefined ? context.query.q : context.query.search;
 
-    console.info(logEntry(context.req, LOG_TYPES.AUDIT, context.query));
-
-    if (noSearchConducted(searchTerm)) {
-        return { props: {} };
-    }
+    if (noSearchConducted(searchTerm)) return { props: {} };
 
     if (emptySearchConducted(searchTerm)) {
         return { props: { searchTerm, isSuccessfulSearch: true, searchResults: [], numberOfMatches: 0 } };
@@ -98,8 +94,10 @@ const getServerSideProps = async (context) => {
     const { isSuccessfulSearch, searchResponseData, searchError } = await searchForCourses(searchTerm);
 
     if (!isSuccessfulSearch) {
-        console.error(logEntry(context.req, LOG_TYPES.ERROR, context.query, { searchError }));
+        logger.error(logEntry(context.req, context.query, null, searchError));
     }
+
+    logger.info(logEntry(context.req, context.query, null));
 
     return {
         props: {
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -96,7 +96,7 @@ const getServerSideProps = async (context) => {
     if (isSuccessfulSearch) {
         logger.info(logEntry(context.req, context.query, null), "User conducted a course search");
     } else {
-        logger.error(logEntry(context.req, context.query, null, searchError), "User failed to conduct a course search");
+        logger.error(logEntry(context.req, context.query, null, searchError), "Course search failed");
     }
 
     return {
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -93,11 +93,11 @@ const getServerSideProps = async (context) => {
 
     const { isSuccessfulSearch, searchResponseData, searchError } = await searchForCourses(searchTerm);
 
-    if (!isSuccessfulSearch) {
-        logger.error(logEntry(context.req, context.query, null, searchError));
+    if (isSuccessfulSearch) {
+        logger.info(logEntry(context.req, context.query, null), "User conducted a course search");
+    } else {
+        logger.error(logEntry(context.req, context.query, null, searchError), "User failed to conduct a course search");
     }
-
-    logger.info(logEntry(context.req, context.query, null));
 
     return {
         props: {
--- a/src/tests/pages/index.test.js
+++ b/src/tests/pages/index.test.js
@@ -221,7 +221,7 @@ describe("getServerSideProps", () => {
                     },
                 },
             },
-            "User failed to conduct a course search"
+            "Course search failed"
         );
     });
 
--- a/src/tests/pages/index.test.js
+++ b/src/tests/pages/index.test.js
@@ -177,7 +177,9 @@ describe("getServerSideProps", () => {
         await getServerSideProps(contextWithSearchTerm);
 
         expect(logger.info).toHaveBeenCalledTimes(1);
-        expect(logger.info).toHaveBeenCalledWith(expect.objectContaining({ parameters: { search: "english" } }));
+        expect(logger.info).toHaveBeenCalledWith(
+            expect.objectContaining({ details: { clientIp: null, parameters: { search: "english" } } })
+        );
     });
 
     it("logs an error when the course search failed", async () => {
@@ -198,24 +200,25 @@ describe("getServerSideProps", () => {
         await getServerSideProps(contextWithSearchTerm);
 
         expect(logger.error).toHaveBeenCalledTimes(1);
-        expect(logger.error).toHaveBeenCalledWith(
-            expect.objectContaining({
+        expect(logger.error).toHaveBeenCalledWith({
+            details: {
+                clientIp: null,
                 parameters: {
                     search: "english",
                 },
-                error: {
-                    message: "Search failed",
-                    searchUrl: "http://foo.bar",
-                    response: {
-                        status: 403,
-                        statusText: "Forbidden",
-                        body: {
-                            message: "Missing authentication token",
-                        },
+            },
+            error: {
+                message: "Search failed",
+                searchUrl: "http://foo.bar",
+                response: {
+                    status: 403,
+                    statusText: "Forbidden",
+                    body: {
+                        message: "Missing authentication token",
                     },
                 },
-            })
-        );
+            },
+        });
     });
 
     it("does not log an error when the course search succeeds", async () => {
--- a/src/tests/pages/index.test.js
+++ b/src/tests/pages/index.test.js
@@ -178,7 +178,8 @@ describe("getServerSideProps", () => {
 
         expect(logger.info).toHaveBeenCalledTimes(1);
         expect(logger.info).toHaveBeenCalledWith(
-            expect.objectContaining({ details: { clientIp: null, parameters: { search: "english" } } })
+            expect.objectContaining({ details: { clientIp: null, parameters: { search: "english" } } }),
+            "User conducted a course search"
         );
     });
 
@@ -200,25 +201,28 @@ describe("getServerSideProps", () => {
         await getServerSideProps(contextWithSearchTerm);
 
         expect(logger.error).toHaveBeenCalledTimes(1);
-        expect(logger.error).toHaveBeenCalledWith({
-            details: {
-                clientIp: null,
-                parameters: {
-                    search: "english",
+        expect(logger.error).toHaveBeenCalledWith(
+            {
+                details: {
+                    clientIp: null,
+                    parameters: {
+                        search: "english",
+                    },
                 },
-            },
-            error: {
-                message: "Search failed",
-                searchUrl: "http://foo.bar",
-                response: {
-                    status: 403,
-                    statusText: "Forbidden",
-                    body: {
-                        message: "Missing authentication token",
+                error: {
+                    message: "Search failed",
+                    searchUrl: "http://foo.bar",
+                    response: {
+                        status: 403,
+                        statusText: "Forbidden",
+                        body: {
+                            message: "Missing authentication token",
+                        },
                     },
                 },
             },
-        });
+            "User failed to conduct a course search"
+        );
     });
 
     it("does not log an error when the course search succeeds", async () => {
--- a/src/tests/pages/index.test.js
+++ b/src/tests/pages/index.test.js
@@ -2,8 +2,10 @@ import { render, fireEvent, screen, within } from "@testing-library/react";
 import { cleanCookies } from "universal-cookie/lib/utils";
 import App, { getServerSideProps } from "../../pages";
 import { searchForCourses } from "../../utils/searchForCourses";
+import { logger } from "../../utils/logger";
 
 jest.mock("../../utils/searchForCourses");
+jest.mock("../../utils/logger");
 
 beforeEach(() => {
     jest.clearAllMocks();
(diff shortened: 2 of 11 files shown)


src/tests/pages/index.test.js
```diff
@@ -198,15 +197,24 @@ describe("getServerSideProps", () => {
...
+                },
+                error: {
+                    message: "Search failed",
+                    searchUrl: "http://foo.bar",
+                    response: {
+                        status: 403,
+                        statusText: "Forbidden",
+                        body: {
+                            message: "Missing authentication token",
+                        },
```
ChristianBeddows, line 214: I really like that we can test that specific values within objects are being logged, this is really cool!

ChristianBeddows (approved):

src/utils/logger.js
```diff
@@ -0,0 +1,34 @@
+const pino = require("pino");
```
FergusMcGlynn, line 1: Minor thing - perhaps we could `import` Pino rather than `require` it, to maintain consistency across the app? Not a big deal though.
```suggestion
import pino from "pino";
```

src/tests/utils/logEntry.test.js
```diff
@@ -1,100 +1,101 @@
...
+        expect(result.error).toEqual(error);
+    });
+
+    it("can parse the http event to find an IP address", () => {
+        const result = logEntry(event, null, null);
+
+        expect(result.clientIp).toEqual("144.32.100.16");
+    });
+
+    it("can handle with no http request", () => {
```
FergusMcGlynn, line 82: Minor thing - this test name doesn't really describe what this test is testing, at least, it's not obvious to me. I guess it's not clear to me what "can handle" means. Same for the following two tests. Could the names describe what behaviour the tests are verifying?

FergusMcGlynn (commented):
Good work! I agree with Christian that the tests are a lot more readable now :) 🥇

src/utils/logEntry.js
```diff
@@ -1,65 +1,45 @@
...
-        const sourcePort = requestHeaders ? requestHeaders["X-Forwarded-Port"] : null;
+const logEntry = (request, parameters, details, error) => {
+    if (error) {
         return {
-            previousIP,
-            sourcePort,
+            clientIp: request ? getClientIp(request) : null,
+            parameters,
+            details,
+            error,
```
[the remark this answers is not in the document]
rmil: Yeah I agree, I'll give them a move, I think with that document I just misinterpreted it wrong. But with regards to the `message` field, the `logger.*` functions accept it as the 2nd argument when called, so I just didn't include it there since I didn't see two much point in being able to do it in two places

src/utils/logEntry.js
```diff
@@ -1,65 +1,45 @@
...
+import { logger } from "./logger";
 
 /**
  *
  * @param {http.IncomingMessage} request
- * @param {String} logType
- * @param {Object} queryParameters
- * @param {Object} additionalDetails Added directly to the log
+ * @param {Object} parameters
+ * @param {Object} details addiotionaly added directly to the log
```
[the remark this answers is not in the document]
rmil: Ah oopsy, should be fixed in 8a2948f

src/utils/logger.js
```diff
@@ -0,0 +1,34 @@
...
+        error: pino.stdSerializers.err,
+    },
+    // Base controls what fields are included in the JSON
+    // object, by default Pino includes fields like pid which
+    // are not very relevant in a serverless context
+    //
+    // We haven't included all fields here as `logEntry` will
+    // add the rest otherwise the keys will duplicate
+    base: {
+        application: "esg-app-course-search",
```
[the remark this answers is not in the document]
rmil: It's a bit of difficult key really, since it does add duplication. But I suppose I would say if we're not including it here, we shouldn't include it in other applications since this key benefits from being present consistently across the applications and it seems odd to not be consistent, so perhaps we need to update that doc?

I think one of the reasons why it's not in the details is since it's a bit meta? Whilst details / parameters I suppose I've viewed them as fields quite relating to the event. But I think I'd be happy to move it also. I think there would also be a slight performance loss since I think when you define it in the base object it only does the operation once? Although I could be wrong

src/utils/searchForCourses.js
```diff
@@ -39,15 +39,17 @@ const searchForCourses = async (searchTerm) => {
     return { isSuccessfulSearch, searchResponseData, searchError };
 };
 
-const shouldRetry = (searchTerm, courseSearchUrl) => (attempt, error, response) => {
+const handleRetry = (searchTerm, courseSearchUrl) => (attempt, error, response) => {
```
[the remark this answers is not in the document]
rmil: I think I changed it since I wasn't too keen on how it read [here](https://github.com/university-of-york/uoy-app-course-search/pull/161/files/f269a0f60698a58410d83b95a95bbe3a21ea7760#diff-f7cf550122384b0eb4e0f60d3c7a48eee46222903f0c82bd505e78fac4b14619L20-R20). But yeah do agree it still isn't perfect. I'm happy to change it back, but what about something like `checkShouldRetry`, or do you think that's a bit too long?

src/utils/logEntry.js
```diff
@@ -1,65 +1,45 @@
...
-const logEntry = (request, logType, queryParameters, additionalDetails) => {
-    const getHeaderInfo = () => {
-        const requestHeaders = request?.headers;
-
-        const ipAddresses = requestHeaders ? requestHeaders["X-Forwarded-For"]?.split(",") : null;
-        const previousIP = ipAddresses?.length > 0 ? ipAddresses[ipAddresses.length - 1].trim() : null;
-
-        const sourcePort = requestHeaders ? requestHeaders["X-Forwarded-Port"] : null;
+const logEntry = (request, parameters, details, error) => {
+    if (error) {
```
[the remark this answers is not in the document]
rmil: Ah yeah that is nicer. I didn't realise that I could do it like that since I thought it would be a bit problematic, but making the change looks like it's quite fine with that. I've updated it now in 01cd00e

src/utils/logger.js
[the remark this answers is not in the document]
FergusMcGlynn: I don't think there's any specific requirement to include `application` in our logs. From my interpretation of the [team's logging structure](https://wiki.york.ac.uk/display/ittechdocs/AWS+Application+logging+fields), the only required fields are `level`, `timestamp`, `message` and `details`. Also that those (plus `error`) are the only fields that should be present at the top level.

As for the potential performance hit - I wouldn't worry about it too much. I doubt it'd be noticeable to an end user :)

src/utils/logger.js
rmil: Ah good catch, that'll be me from copying from courses-api :sweat_smile:. I've updated this in 4275374 but I've also taken it a step further and updated the export style as well so it's consistent

src/utils/searchForCourses.js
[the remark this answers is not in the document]
FergusMcGlynn: I think I still prefer `shouldRetry`. There is a convention that boolean functions [should be named with a yes/no question](https://stackoverflow.com/a/1370847). It's not universally adopted, but I think it leads to more readable code.

src/tests/utils/logEntry.test.js
rmil: I've updated the name in 38f2365, hopefully that is more clear. Also the other two tests I've removed since they were a bit redundant after removing the logic from `logEntry` with regards to error handling

src/utils/searchForCourses.js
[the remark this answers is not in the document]
rmil: Fixed in db15dbe

src/utils/logEntry.js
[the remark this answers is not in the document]
rmil: Fixed in c9ae138

src/utils/logger.js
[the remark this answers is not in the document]
rmil: Fixed in c1ec485 & 5bdb356

src/utils/logEntry.js
```diff
@@ -1,65 +1,44 @@
...
-            attempt,
-            error: error?.message,
-            response: JSON.stringify(response),
-        })
+    logger.warn(
+        logEntry(
+            null,
+            { search: searchTerm },
+            {
+                message: "Request failed, retrying",
```
FergusMcGlynn, line 35: Should `message` to be at the top level of the log entry, rather than underneath `details`?

src/utils/logEntry.js
```diff
@@ -1,65 +1,44 @@
...
-
-        const ipAddresses = requestHeaders ? requestHeaders["X-Forwarded-For"]?.split(",") : null;
-        const previousIP = ipAddresses?.length > 0 ? ipAddresses[ipAddresses.length - 1].trim() : null;
-
-        const sourcePort = requestHeaders ? requestHeaders["X-Forwarded-Port"] : null;
-        return {
-            previousIP,
-            sourcePort,
-        };
+const logEntry = (request, parameters, details, error) => {
```
FergusMcGlynn, line 12: Could this function possibly be simplified to this?
```
const logEntry = (request, parameters, details, error) => {
    const clientIp = request ? getClientIp(request) : null;
    return {
        error,
        details: { ...details, parameters, clientIp },
    };
};
```
Or even
```
const logEntry = (request, parameters, details, error) => ({
    error,
    details: { ...details, parameters, clientIp: request ? getClientIp(request) : null },
});
```

src/utils/logEntry.js
```diff
@@ -1,65 +1,44 @@
...
 
 /**
  *
  * @param {http.IncomingMessage} request
- * @param {String} logType
- * @param {Object} queryParameters
- * @param {Object} additionalDetails Added directly to the log
+ * @param {Object} parameters
+ * @param {Object} details additionally added directly to the log
+ * @param {Error} error
```
FergusMcGlynn, line 9: The `error` parameter, at least when being passed in from `index.js`, is not of type `Error` but of type `Object`. It's just an object containing fields relevant to the error that occurred, but it's not actually a proper `Error`.

src/pages/index.js
```diff
@@ -98,9 +94,11 @@ const getServerSideProps = async (context) => {
     const { isSuccessfulSearch, searchResponseData, searchError } = await searchForCourses(searchTerm);
 
     if (!isSuccessfulSearch) {
-        console.error(logEntry(context.req, LOG_TYPES.ERROR, context.query, { searchError }));
+        logger.error(logEntry(context.req, context.query, null, searchError));
     }
 
+    logger.info(logEntry(context.req, context.query, null));
```
FergusMcGlynn, line 100: Since `logEntry` takes 4 parameters, should we include the 4th parameter in this function call (presumably it'd be `null`)?

src/utils/logEntry.js
FergusMcGlynn: In which case - might this make a log entry unparseable by the CloudWatch Monitor? I think it expects that if there is an `error` field in the log entry, it'd contain `type`, `message`, `details` and optionally `stack`. Here, I think only `message` and  sometimes `details` would be present.

src/pages/index.js
```diff
@@ -98,9 +94,11 @@ const getServerSideProps = async (context) => {
     const { isSuccessfulSearch, searchResponseData, searchError } = await searchForCourses(searchTerm);
 
     if (!isSuccessfulSearch) {
-        console.error(logEntry(context.req, LOG_TYPES.ERROR, context.query, { searchError }));
+        logger.error(logEntry(context.req, context.query, null, searchError));
```
FergusMcGlynn, line 97: It looks to me like neither this call to `logger.error`, nor the call to `logger.info` a few lines below, will result in a log entry containing a `message` field at the top level. Perhaps we need to add this in? For example:
```
logger.error(logEntry(context.req, context.query, null, searchError), searchError.message);
```
and
```
logger.info(logEntry(context.req, context.query, null, null), "User conducted a course search");
```

src/utils/logEntry.js
rmil: Should be fixed in 5287d11

src/utils/logEntry.js
rmil: Should be fixed in 5bdb356

src/pages/index.js
rmil: I'm happy to leave it as-is since there isn't much difference between the two as far as I'm aware right now

src/pages/index.js
rmil: I've popped this together, 6f374ed, I've made the error one instead say `"User failed to conduct a course search"`. When templated the `message` property on `searchError` should be included, which would then have the same message appear twice I suppose?

src/utils/logEntry.js
rmil: Quite true, I've rustled this up: 16d8095, which adds a type field

src/utils/searchForCourses.js
```diff
@@ -12,7 +12,11 @@ const searchForCourses = async (searchTerm) => {
 
     let isSuccessfulSearch;
     let searchResponseData = { numberOfMatches: 0, results: [] };
-    let searchError = { message: "Failed to fetch results from Courses API", searchUrl: courseSearchUrl };
+    let searchError = {
+        message: "Failed to fetch results from Courses API",
+        type: "SearchError",
+        searchUrl: courseSearchUrl,
```
FergusMcGlynn, line 18: I like the addition of `type` here - that should help the CloudWatch Monitor parse these log entries :)

Looking at [the CloudWatch Monitor validation](https://github.com/university-of-york/esg-app-tl-cloudwatch-monitor/blob/dev/src/error-notification-lambda/utils/parseCloudWatchLogEvent.ts#L16) though, I think this `SearchError` object will need to always have a `details` field if it's going to be nicely templated. In which case, perhaps we could move `searchUrl` under a `details` field, e.g.
```
let searchError = {
        message: "Failed to fetch results from Courses API",
        type: "SearchError",
        details: {
            searchUrl: courseSearchUrl,
        }
}
```
and then later on in this function (line 32 currently)
```
        searchError.details.response = {
```
Then `details` should always be present, meaning the log should always be parse-able, and it will contain all the useful information. What do you think?

src/pages/index.js
```diff
@@ -85,20 +85,18 @@ App.propTypes = {
...
     }
 
     const { isSuccessfulSearch, searchResponseData, searchError } = await searchForCourses(searchTerm);
 
-    if (!isSuccessfulSearch) {
-        console.error(logEntry(context.req, LOG_TYPES.ERROR, context.query, { searchError }));
+    if (isSuccessfulSearch) {
+        logger.info(logEntry(context.req, context.query, null), "User conducted a course search");
+    } else {
+        logger.error(logEntry(context.req, context.query, null, searchError), "User failed to conduct a course search");
```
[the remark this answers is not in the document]
rmil: Ah yeah! I've updated in it here: 3e45fca

src/utils/searchForCourses.js
rmil: Hopefully resolved in eeeba8b

FergusMcGlynn (approved):

